PR #6, opened 2020-05-06
Add govuk-frontend to dummy app

rjlynch:
Pulls in govuk-frontend so components can be tried out in the dummy app.

Testing:
Generate a new controller and view in the demo app and try out some of the components.

Screenshot 2020-05-06 at 09 48 06

We might want to consider using the dummy app as a catalogue for the various components.


tvararu:
Can we use [ViewComponent previews](https://github.com/github/view_component#previewing-components) for this?

rjlynch:
> Can we use [ViewComponent previews](https://github.com/github/view_component#previewing-components) for this?

Good idea, that would work better for cataloguing the available components.
I've tested it out with previews in the dummy app and they work fine.

rjlynch:
> Awesome, worked fine for me.
> 
> Think it's worth adding a controller and view with a demo of the components and setting it as the app's root?

Yeah I think so, or maybe root to the components preview index 🤔

PR #8, opened 2020-05-06
Add a phase banner component

peteryates:
This component adds the [Phase Banner](https://design-system.service.gov.uk/components/phase-banner/) component. The intent is for it to be used in a simple manner with a `text:` argument and if more complicated content is needed (ie. text containing an anchor tag or italicised text) should support HTML via a block:

```erb
<%= render(GovukComponent::PhaseBanner.new(tag: 'Lorem')) do |phase_banner| %>
  <%= phase_banner.with(:html) do %>
     Ipsum dolor sit <strong>amet</strong> consectetur 
  <% end %>
<% end %>
```
vs

```erb
<%= render(GovukComponent::PhaseBanner.new(tag: 'Lorem', text: 'Ipsum dolor sit amet')) %>
```


peteryates:
So this is _more or less_ ready. The tests for handling blocks is a bit of a pain (see 1f2240195fd1), this is something that's not yet handled by the gem. Not being able to test block content in unit tests seems like a shortfall of the gem. Hopefully it'll be fixed by the time of Rails 6.1.

Should the block tests be split and moved to an integration test? That would mean tests are split up and I'm not sure that testing everything in an test is a good move.

peteryates:
I agree @tvararu, it just means components that take blocks can't be tested yet.

peteryates:
Example with HTML passed into the block:

```erb
<%= render GovukComponent::PhaseBanner.new(tag: 'hi') do %>
  the quick brown <em>fox</em> jumped over the <strong>lazy</strong> dog
<% end %>
```

Screenshot from 2020-05-12 11-22-12

PR #10, opened 2020-05-12
Move specs out of panel dir

rjlynch:
Moves the component specs to `spec/components/govuk_component`
The component specs were nested under the panel dir `spec/components/panel/govuk_component` I'm guessing this wasn't intentional!


peteryates:
:man_facepalming:

PR #12, opened 2020-05-13
Add breadcrumbs component

peteryates:
Add a basic breadcrumbs component. It takes a hash `breadcrumbs` in the format `{ link => target }`

```ruby
@breadcrumbs = {
  'Level one'   => '/level-one/',
  'Level two'   => '/level-one/level-two/',
  'Level three' => nil   # don't link to the current page
}
```

```erb
render GovukComponent::Breadcrumbs.new(breadcrumbs: @breadcrumbs)
```


tvararu:
👍

PR #13, opened 2020-05-14
Add Back link component

aliuk2012:
This PR adds the govuk-frontend back link component:

Guidance and design:
https://design-system.service.gov.uk/components/back-link/

At the moment this component allows developers to pass in
- custom link text
- href
- css classes (optional)
- html attributes (optional)

screencapture-localhost-3000-2020-05-14-22_56_24

### DIFF
--- a/app/components/govuk_component/back_link.html.erb
+++ b/app/components/govuk_component/back_link.html.erb
@@ -1,3 +1,1 @@
-<a href="<%= @href %>" class="govuk-back-link<% if @classes.present? %> <%= @classes%><% end %>">
-  <%= @text %>
-</a>
+<%= link_to @text, @href, @options  %>
--- a/app/components/govuk_component/back_link.html.erb
+++ b/app/components/govuk_component/back_link.html.erb
@@ -1,0 +1,3 @@
+<a href="<%= @href %>" class="govuk-back-link<% if @classes.present? %> <%= @classes%><% end %>">
+  <%= @text %>
+</a>
--- a/app/components/govuk_component/back_link.rb
+++ b/app/components/govuk_component/back_link.rb
@@ -1,0 +1,9 @@
+class GovukComponent::BackLink < ViewComponent::Base
+  attr_accessor :text, :href
+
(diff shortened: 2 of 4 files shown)


app/components/govuk_component/back_link.html.erb
```diff
@@ -0,0 +1,3 @@
+<a href="<%= @href %>" class="govuk-back-link<% if @classes.present? %> <%= @classes%><% end %>">
```
aliuk2012, line 1: I would like to extend this to allow additional html attributes.

aliuk2012: Additional html attributes can now be added by using the optional `attributes` params

app/components/govuk_component/back_link.rb
```diff
@@ -0,0 +1,13 @@
+class GovukComponent::BackLink < ViewComponent::Base
+  attr_accessor :text, :href, :options
+
+  def initialize(text:, href:, classes: nil, attributes: nil)
+    @text = text
+    @href = href
+    @classes = "govuk-back-link"
+    @classes << " #{classes}" if classes.present?
```
peteryates, line 8: Would it make sense to be able to pass in an array of classes?

tvararu: I think either approach can work fine, but possibly passing it as a string is less opinionated. I think most component libraries I've worked with expect strings, because that's what the core HTML elements themselves expect. Leave it to the caller to figure out how to make that string based on their requirements.

tvararu: What I like about either strings or array of strings is that they both should work with https://purgecss.com; as long as that is maintained, we should be good with either!

aliuk2012: Like @tvararu, I think most component libraries expect strings. I suppose if we really want to pass arrays we could check whether `classes` is an array or string and then build up the string internally. 

Either way I think we should be consistent with our approach because I think alot (if not all) components should support custom css classes like `govuk-frontend` does.

PR #14, opened 2020-05-14
Adds demonstration page

rjlynch:
I was thinking we could reuse the dummy app's `demos#show` page as online documentation. If we generate a static version of `demos#show` with assets inlined and place it in a top level `docs` directory we could enable github pages for this repo and it _should_ pick it up.
I think this approach is simpler than pushing the dummy app to heroku and getting that working.

I've added a rake task to generate the static page but it's pretty brittle imo. There's probably a nicer/more robust way to inline the assets but I doubt the dummy app will change much, so might be fine ¯\_(ツ)_/¯. Thoughts?

Testing: 
cd into the dummy apps directory and run `bin/rake generate_examples_page`
open `docs/index.html` it should look correct!

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,0 +1,10522 @@
+<!DOCTYPE html>
+<html>
+  <head>
+<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    <title>Dummy</title>
+    <meta name="csrf-param" content="authenticity_token">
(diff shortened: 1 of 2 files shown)


peteryates:
This would be great. Having a quick reference with examples is worth its weight in gold.

peteryates:
One other observation - should `docs/index.html` be in the `.gigignore`?

rjlynch:
> One other observation - should `docs/index.html` be in the `.gitignore`?

No, it needs to be in the repo so github pages can pick it up. We could generate it with github actions but I think for the minute manually running the rake task will suffice

Issue #3, opened 2020-05-05
Components in scope for this gem


tijmenb:
For visibility, @peteryates, @tvararu and I had a discussion on the DfE Slack about the scope for this library.

The consensus (I think) is that this library lives along side other projects that enable developers to build applications using the GOV.UK Design System. For forms, there's [govuk_design_system_formbuilder](https://github.com/DFE-Digital/govuk_design_system_formbuilder), which means that form elements are out of scope for this gem. We suspect that some GOV.UK Design System components like tables might be too complex to be built nicely using ViewComponent, and would benefit from a custom DSL.

This is a quick list of components and whether or not they're in scope. Feel free to edit & comment.

Component | In scope?  
-- | --
Accordion | Too complex? Maybe in 2 components
Back link | Yes, in scope for this gem
Breadcrumbs | Yes, in scope for this gem
Button | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Character count | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Checkboxes | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Date input | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Details | Yes, in scope for this gem
Error message | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Error summary | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Fieldset | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
File upload | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Footer | Yes, in scope for this gem
Header | Yes, in scope for this gem
Inset text | Yes, in scope for this gem
Panel | Yes, in scope for this gem
Phase banner | Yes, in scope for this gem
Radios | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Select | No, part of [form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder)
Skip link | Yes, in scope for this gem
Summary list | Yes, in scope for this gem
Table | No - too complex. We may create a different gem.
Tabs | Yes, in scope for this gem
Tag | Yes, in scope for this gem
[...]

PR #15, opened 2020-05-17
Add basic govuk_link_to helper and test

peteryates:
This is a bit of an experiment related to the suggestion made in the comments of #3, it adds a `app/helpers` directory and an initial `#govuk_link_to` helper that's a wrapper of Rails' `#link_to`

* [x] Add a proof of concept helper
* [x] Ensure it works with the dummy app
Screenshot from 2020-05-17 12-33-49

* [x] Add a test (and make it pass!)

Testing currently doesn't work
* in the submitted state we get `undefined method `link_to'`
* adding `include ActionView::Helpers::UrlHelper` to the top of `govuk_link_helper_spec.rb` or `spec_helper.rb` makes the individual file pass but when running the suite we get `cyclic include detected`

Not quite sure how to make it play ball with RSpec.


aliuk2012:
Can you pass custom css classes?

peteryates:
Just finishing off that functionality @aliuk2012, ensuring it works for strings and arrays

peteryates:
@aliuk2012 so yeah, having had a think any custom classes will override the default. Ensuring safe merging made it a bit too complex. We can revisit this behaviour if needs be

Issue #32, opened 2020-05-26
For Discussion: CSS classes option consistency 


aliuk2012:
## Background
We need components to have a consistent API otherwise it will make it difficult to implement without looking at documentation. It will also make it confusing to know which method to use for each component.

Eg.

- Backlink uses a named parameter called classes and allows you to pass in a string. It does use Rails built in `link_to` method . https://github.com/DFE-Digital/govuk-components/blob/master/app/components/govuk_component/back_link.rb#L7-L10
- warning component uses an array of of classes. https://github.com/DFE-Digital/govuk-components/blob/master/app/components/govuk_component/warning.html.erb#L1

## Proposal 

I feel that we should support a string of class names as default as most frameworks support this. In addition if its really necessary then we should also support an array of class names and let the component convert the array to a string. BTH, when I see classes as an array I automatically start wondering if theres a priority in which the classes need to be in or whether the order of it matters (which it shouldn't). This is another reason why I think string of class names is easer to use because these thoughts or concerns are not brought about because of it being an array, it is just a string after all when its rendered.

🤔 Should we have a helper method which accepts string/array of classes and converts it into a string of class names for components to reuse? This could be used in other components which don't currently support custom css class names.

Any other thoughts or suggestions from others on the subject?

peteryates:
My plan for this was to get an initial draft of the components in and then look at ensuring the API is consistent via something like [form builder's traits](https://github.com/DFE-Digital/govuk_design_system_formbuilder/tree/master/lib/govuk_design_system_formbuilder/traits) and making them inherit from a single [base class](https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/master/lib/govuk_design_system_formbuilder/base.rb#L6); then we can make these decisions in a single place and enforce uniformity.

I don't mind too much whether the `classes` argument takes a `String`, `Array` but I'm leaning towards having it accept both. Internally, combining arrays feels much cleaner than appending strings and there are patterns [which work nicely in the form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/master/lib/govuk_design_system_formbuilder/containers/radios.rb#L21) that we can use for inspiration.

When this is dealt with in a single place we can refactor and hone it until everyone's happy. I definitely think that standardising should be the next step once the components are implemented, although there's no reason we can't start that now.

peteryates:
[...]

PR #33, opened 2020-06-05
Use a common base class to standardise behaviour

peteryates:
Some behaviour is likely to be shared across many components, such as the ability to take custom classes. This approach introduces `GovukComponent::Base` which handles this; it has been applied to `Warning` and `BackLink`.

Refs #32


peteryates:
The requested review is more for feedback and discussion on the approach than reviewing the PR as a whole. It can easily be extended if everyone's happy with it

PR #67, opened 2020-10-08
Fix Gem::Specification homepage

benilovj:
When this parameter isn't correct, the govuk-components link in Dependabot PRs isn't pointing to the right place:

image

(is this also why we're not seeing the commit and version history?)


peteryates (approved):
Thank you!

benilovj:
(I don't think I have merge permissions on this repo...)

PR #90, opened 2021-01-08
Support notification banner content via blocks

peteryates:
This gives the user the option between using structured arguments and just passing in text and link information vs providing the heading with a chunk of HTML that will be rendered within the `p.govuk-notification-banner__heading`.

Screenshot from 2021-01-08 09-24-35


peteryates:
This could possibly be enhanced further with a specific link helper that adds [the appropriate classes for notification banner header links](https://design-system.service.gov.uk/components/notification-banner/).

Issue #53, opened 2020-09-08
Table component


peteryates:
Tables might be too complicated to implement using components (#18) as they could potentially need two 'layers' of slots - one for rows and one within it for cells. More investigation is needed

* [ ] Is it possible to nest slots? Can we use this technique to build a table?
* [ ] Would a 'simple' table component work as an alternative, pass in a array of arrays or array of hashes (like `CSV`) and it builds a table itself? `= govuk_table(@teachers)`

cpjmcquillan:
My (teaching-vacancies) original implementation of the summary list component took an array of hashes instead of slots and it worked quite well. I'm about to implement a table component that hopefully includes some sorting bells and whistles - will see how it turns out and link to any findings in here!

peteryates:
Did you make any progress on your table component @cpjmcquillan?

cpjmcquillan:
> Did you make any progress on your table component @cpjmcquillan?

Only just seen this! Eventually we moved away from tables completely since they're not great on smaller screens

peteryates:
Ah, I had a first go at them in #100 but realised my approach was flawed.

Agree on them being annoying on mobile, you know it's bad when even `psql` handles smaller screens better 😞

cpjmcquillan:
> Ah, I had a first go at them in #100 but realised my approach was flawed.
> 
> Agree on them being annoying on mobile, you know it's bad when even `psql` handles smaller screens better 😞

On our service we've started using a concept of cards (which work really nicely on smaller screens). Maybe they'll make their way into the design system one day 🙏

PR #100, opened 2021-01-30
Add basic table component and specs

peteryates:
This is a first stab at a generic, simple table component.

It can be generated with:

* a multidimensional array,
* an array of hashes or,
* an array of objects.

Headings must be provided and aren't inferred.

It's not meant to cover all of the possible table functionality but will
eventually be able to:

* right align columns
* make the first (or perhaps any) column a header column
* add captions

Refs #3, #53


peteryates:
Closing to rewrite with Slots v2 in mind.

PR #108, opened 2021-02-17
Cookie banner component

cpjmcquillan:
For some reason the component doesn't look quite so beautiful in the dummy app as it does [in the design system](https://design-system.service.gov.uk/components/cookie-banner/) for reasons unknown. If anybody has some ideas about why, give me a shout. 

Also not sure on PR etiquette in this repo, but I've tried to follow patterns properly.

Let me know any thoughts you have!

## Changes in this PR
- Update govuk-frontend in the dummy app
- Add GovukComponent::CookieBanner view component

## Screenshots
image

### DIFF
--- a/app/components/govuk_component/cookie_banner.html.erb
+++ b/app/components/govuk_component/cookie_banner.html.erb
@@ -1,4 +1,4 @@
-<%= tag.div(class: classes, **html_attributes) do %>
+<%= tag.div(class: classes, role: "region", aria: { label: aria_label }, **html_attributes) do %>
   <div class="govuk-cookie-banner__message govuk-width-container">
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
--- a/app/components/govuk_component/cookie_banner.html.erb
+++ b/app/components/govuk_component/cookie_banner.html.erb
@@ -1,0 +1,19 @@
+<%= tag.div(class: classes, **html_attributes) do %>
+  <div class="govuk-cookie-banner__message govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h2 class="govuk-cookie-banner__heading govuk-heading-m">
+          <%= title %>
+        </h2>
+
+        <div class="govuk-cookie-banner__content">
+          <%= body %>
+        </div>
+      </div>
+    </div>
+
+    <div class="govuk-button-group">
+      <%= actions %>
+    </div>
+  </div>
+<% end %>
--- a/app/components/govuk_component/cookie_banner.html.erb
+++ b/app/components/govuk_component/cookie_banner.html.erb
@@ -2,9 +2,11 @@
   <div class="govuk-cookie-banner__message govuk-width-container">
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
-        <h2 class="govuk-cookie-banner__heading govuk-heading-m">
-          <%= title %>
-        </h2>
+        <% if title.present? %>
+          <h2 class="govuk-cookie-banner__heading govuk-heading-m">
+            <%= title %>
+          </h2>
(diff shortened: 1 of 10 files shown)


peteryates:
Awesome! Yeah, there are a couple of components that don't render perfectly on the dummy page, looks spot-on in an action page though. Only minor thing left is to add it to the table in the `README` and we're good to go 👍🏽

frankieroberto:
Could we try and follow the API conventions of the Nunjucks macro a bit more closely? Would make converting from prototypes to production code a bit easier.

For example, using `heading` instead of `title` and `html` instead of `body`?

Would also be good to support the `hidden` boolean too, as that might be needed by client-side implementations.

app/components/govuk_component/cookie_banner.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= tag.div(class: classes, **html_attributes) do %>
+  <div class="govuk-cookie-banner__message govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h2 class="govuk-cookie-banner__heading govuk-heading-m">
+          <%= title %>
+        </h2>
```
frankieroberto, line 7: Need a condition here, as title/heading isn't required.

```suggestion
        <% if title %>
          <h2 class="govuk-cookie-banner__heading govuk-heading-m">
            <%= title %>
          </h2>
        <% end %>  
```

app/components/govuk_component/cookie_banner.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+        <div class="govuk-cookie-banner__content">
+          <%= body %>
+        </div>
+      </div>
+    </div>
+
+    <div class="govuk-button-group">
+      <%= actions %>
+    </div>
```
frankieroberto, line 17: Need a condition here, as actions aren't required (eg if only essential cookies are used).

```suggestion
    <% if actions && actions.length > 0 %>
      <div class="govuk-button-group">
        <%= actions %>
      </div>
    <% end %>
```

app/components/govuk_component/cookie_banner.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= tag.div(class: classes, **html_attributes) do %>
+  <div class="govuk-cookie-banner__message govuk-width-container">
```
frankieroberto, line 2: It'd be good to have a way to inject additional classes here...

app/components/govuk_component/cookie_banner.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= tag.div(class: classes, **html_attributes) do %>
```
frankieroberto, line 1: I think we're missing the `role="region"` and `aria-label` attributes here? See [Nunjucks version](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/cookie-banner/template.njk#L3).

frankieroberto (changes requested):
Thanks for working on this!

peteryates:
Thanks for the review @frankieroberto, I've made the title and actions optional and allowed for a customisable `aria-label`

For the time being I think we should stick with `title` and `body` as they're already used in various other components.

I'm not keen on the `html` param name used by the nunjucks repos as it's very general, but if there's a call for this library to be more in-line with them I'd be happy to migrate it that way at the point of a major release.

peteryates:
I just force pushed without the commit that allows for no actions. I think at least one action is necessary to be able to dismiss the banner. In the case of essential non-tracking cookies only I think not showing the banner is a more sensible move.

frankieroberto:
@peteryates thanks!

With all these ports of `govuk-frontend` there's a bit of a trade-off between making them more similar to the Nunjucks version vs making them more similar to the conventions of the language they're being ported into. It's a tricky balance!

PR #115, opened 2021-03-02
Add `expanded` option to accordion slots

stevehook:
I wanted to use the Accordion component to replace some long-hand markup in Find Teacher Training (https://github.com/DFE-Digital/find-teacher-training/blob/13f583cb053a5a7687465fccbb66200261b955f4/app/views/result_filters/subject/_form.html.erb#L10) and found that there was no way to specify an initially expanded section (slot). This PR is an attempt to add the option to the Accordion component's Slot sub-component.

I don't think this can be done simply by adding CSS class due the `aria` attribute needed on the heading so I've added an explicit `expanded` option here. There may be a simpler way!

One source of confusion (for me) was some JavaScript included in the design system that persists the expanded state of accordions in local storage. So if you've already viewed a page, toggled the expanded state and then revisit it whatever is in local storage overrides whatever the markup specifies.

### DIFF
--- a/app/components/govuk_component/accordion.html.erb
+++ b/app/components/govuk_component/accordion.html.erb
@@ -3,7 +3,7 @@
     <%= tag.div(id: section.id(suffix: 'section'), class: section.classes, **section.html_attributes) do %>
       <div class="govuk-accordion__section-header">
         <h2 class="govuk-accordion__section-heading">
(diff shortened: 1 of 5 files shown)


peteryates (approved):
Looks good to me 👍🏽

PR #124, opened 2021-03-25
Avoid rails therefore mimemagic as a dependency

asmega:
## Issue

- There are licensing issues with mimemagic
- Further discussion can be found here at https://github.com/rails/rails/issues/41750

## The fix

- Remove rails as a dependency and only include the dependencies we need so mimemagic is excluded


peteryates (approved):
Looks good for me, tests all pass locally 😎

Issue #137, opened 2021-03-30
Start now button: Incorrect name?


paulrobertlloyd:
Now this is _really_ pernickety, but… what’s called the ‘Start now button’ component in this gem, is commonly referred to as the ‘Start button’. See: https://design-system.service.gov.uk/components/button/#start-buttons (Also, the class is `govuk-button--start`).

Suppose it’s too late to rename this component… or at least until there’s a breaking change?

frankieroberto:
Arguably it could be an option passed to the `govuk_link_to` or `govuk_button_to` helpers instead? Or we create a new `Button` component...

Trade-offs between more closely following the `govuk-frontend` macros vs the Rails conventions...!

peteryates:
I did a quick GitHub search to see if the component was in use anywhere and there are a few - I'll hold off the renaming until the next major version and leave this issue as a reminder.

frankieroberto:
In _theory_ we shouldn't be using the Start now button in services, as it should only be on GOV.UK start pages... ;-)

Issue #136, opened 2021-03-30
Phase banner: Allow customisation of the phase tag


paulrobertlloyd:
The [equivalent Nunjucks macro for this component](https://design-system.service.gov.uk/components/phase-banner/) allows for a greater deal of customisation, given that it accepts all the arguments for `tag` that are available to the [tag component](https://design-system.service.gov.uk/components/tag/). It would be great if this component behaved in a similar way.

On Apply, we change the colour of the tag based on the environment (production, staging, QA, etc.). However, it’s not possible to achieve the same with the phase banner component as it currently stands in this gem.

Issue #135, opened 2021-03-30
Panel: Body should be an optional value


paulrobertlloyd:
Whether panels should contain body content or not is a separate question, but we use the panel component in a few places where we only show a title. Right now the panel component requires a value for `body`, meaning we need to set a `nil` value for `body` in these instances:

```erb
<%= govuk_panel(
  title: 'You’ve successfully signed the data sharing agreement',
  body: nil,
) %>
```

Ideally the `body` argument (and possibly even the `title` argument) would be optional.

Issue #134, opened 2021-03-30
Inset text: Accept complex content


paulrobertlloyd:
The inset text component can only accept a value for `text`. This is fine for one line of text, but requires significant work arounds when multiple lines and other elements need to be shown within this component. For example, on Apply we have the following:

```html
<div class="govuk-inset-text">
  <% if no_viable_references? %>
    <p class="govuk-body">You need 2 references before you can submit your application.</p>
    <%= govuk_button_link_to 'Add a referee', candidate_interface_references_start_path %>
  <% elsif one_viable_reference? %>
    <p class="govuk-body">You need 2 references before you can submit your application.</p>
    <%= govuk_button_link_to 'Add a second referee', candidate_interface_references_start_path %>
  <% else %>
    <p class="govuk-body">You can add more referees to increase the chances of getting 2 references quickly.</p>
    <p class="govuk-body">We’ll cancel any remaining requests when you’ve received 2 references.</p>
    <%= govuk_button_link_to 'Add another referee', candidate_interface_references_start_path, class: 'govuk-button--secondary' %>
  <% end %>
</div>
```

It’d be nice if the inset text component could accept either block content or use a slot.

PR #138, opened 2021-03-31
General component tweaks and improvements

peteryates:
This PR contains some general improvements to several components suggested by
@paulrobertlloyd.

* [x] panel component (c261410bc87 / #135)
* [x] inset text (c6917a8 / #134 )
* [x] header (24907b4 / #133) 
* [x] phase banner (6b3b66fe9c / #136)
* [x] ~start button~ (holding off for the time being #137)

### DIFF
--- a/app/components/govuk_component/inset_text.rb
+++ b/app/components/govuk_component/inset_text.rb
@@ -1,14 +1,18 @@
 class GovukComponent::InsetText < GovukComponent::Base
   attr_accessor :text
 
-  def initialize(text:, classes: [], html_attributes: {})
+  def initialize(text: nil, classes: [], html_attributes: {})
     super(classes: classes, html_attributes: html_attributes)
 
     @text = text
   end
 
   def call
-    tag.div(class: classes, **html_attributes) { @text }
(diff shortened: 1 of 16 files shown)


cpjmcquillan (approved):

app/components/govuk_component/inset_text.rb
```diff
@@ -1,14 +1,18 @@
...
-  def initialize(text:, classes: [], html_attributes: {})
+  def initialize(text: nil, classes: [], html_attributes: {})
     super(classes: classes, html_attributes: html_attributes)
 
     @text = text
   end
 
   def call
-    tag.div(class: classes, **html_attributes) { @text }
+    tag.div(class: classes, **html_attributes) { @text || content }
```
frankieroberto, line 11: If both a block and `text` are specified, would it be more conventional to prioritise the block? (I think this is what the [rails `content_tag` does](https://github.com/rails/rails/blob/main/actionview/lib/action_view/helpers/tag_helper.rb#L320-L327)). Alternatively it could throw an exception?

PR #147, opened 2021-04-28
Resolve `with_content_areas` view_component deprecation warning

despo:
> DEPRECATION WARNING: `with_content_areas` is deprecated and will be removed in ViewComponent v3.0.0.
Use slots (https://viewcomponent.org/guide/slots.html) instead. (called from <top (required)> at .../apply-for-teacher-training/config/environment.rb:8)

This PR also fixes a non working spec, as testing components using a `within` block does not appear to be supported and results in a false positive.


peteryates (approved):
Nice!

PR #151, opened 2021-05-03
Fix header component

paulrobertlloyd:
There are still some outstanding issues with the header component, which this PR attempts to address.

* Fixes trailing whitespace around the  ‘Menu’ text. This space is most noticeable when hovering over the button when this space is underlined
* Fixes the menu button not showing or hiding the navigation menu
* Add a new `navigation_label` param. This matches the `navigationLabel` setting available in the design system component: 

  > Text for the `aria-label` attribute of the navigation. Defaults to "Navigation menu".

* Changes how `menu_button_label` is applied, to match the corresponding `menuButtonLabel` setting used by the design system component (**breaking change**):
 
  > Text for the `aria-label` attribute of the button that toggles the navigation. Defaults to "Show or hide navigation menu".

* * *

I’ve tied to update the corresponding tests, but there was no documentation on how to run tests on this project. I tried `cd spec/dummy` and running `bin/rake spec`, but not sure this was right.

### DIFF
--- a/app/components/govuk_component/header.html.erb
+++ b/app/components/govuk_component/header.html.erb
@@ -32,7 +32,7 @@
       <% if items.any? %>
         <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide navigation menu"><%= menu_button_label %></button>
         <nav>
-          <%= tag.ul(class: navigation_classes, aria: { label: "Top Level Navigation" }) do %>
+          <%= tag.ul(class: navigation_classes, id: "navigation", aria: { label: "Top Level Navigation" }) do %>
             <% items.each do |item| %>
               <%= tag.li(class: item.classes.append(item.active_class), **item.html_attributes) do %>
                 <% if item.link? %>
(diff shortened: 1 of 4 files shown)


app/components/govuk_component/header.html.erb
```diff
@@ -30,11 +30,9 @@
       <% end %>
 
       <% if items.any? %>
-        <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide navigation menu">
-          <%= menu_button_label %>
-        </button>
+        <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="<%= menu_button_label %>">Menu</button>
```
peteryates, line 33: I think I must've misunderstood your previous request. I thought `menu_button_label` would control the content of the button on mobile:

Screenshot from 2021-05-04 18-00-49

Should it be renamed to `menu_button_aria_label` perhaps?

peteryates (commented):
Looks great overall - just a question on the param naming and the actual button text.

app/components/govuk_component/header.html.erb
paulrobertlloyd: > Should it be renamed to menu_button_aria_label perhaps?

I don’t think so. I really do believe that, where possible, this gem should share the same names and values as those used by the design system Nunjucks components. Unfortunate that in this case this is a breaking change, but maybe still early enough to mitigate?

peteryates (approved):

PR #157, opened 2021-05-11
WIP: Update slots usage

cpjmcquillan:
Slots V1 API is now deprecated so we should update usage. https://viewcomponent.org/guide/slots.html

I've made a start here, but am having some issues that I haven't gotten to the bottom of to do with delegate slots and content blocks.


peteryates:
Should be solved by #160 ☝🏽

PR #164, opened 2021-05-24
Update the back link and breadcrumb components

peteryates:
Upgrade both the backlinks and breadcrumbs components and make them, and their associated specs, match the new style.


spec/components/govuk_component/back_link_component_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  let(:text) { 'Department for Education' }
+  let(:href) { 'https://www.gov.uk/government/organisations/department-for-education' }
+  let(:kwargs) { { text: text, href: href } }
+  let(:component_css_class) { 'govuk-back-link' }
+
+  subject! { render_inline(GovukComponent::BackLinkComponent.new(**kwargs)) }
+
+  specify 'renders a link with the right href and text' do
+    expect(page).to have_css('a', text: text, class: component_css_class)
```
cpjmcquillan, line 14: Looks like you're missing the href! I _think_ that should work...

```suggestion
    expect(page).to have_css('a[href="#{href}"]', text: text, class: component_css_class)
```

cpjmcquillan: Although based on the breadcrumbs spec, I expect you might use `have_tag` instead

cpjmcquillan (approved):

spec/components/govuk_component/back_link_component_spec.rb
peteryates: Ah thanks for that, forgot to switch this test over to RSpec HTML matcher syntax too.

PR #170, opened 2021-05-27
Version 2.0.0 DetailsComponent

cpjmcquillan:
- Remove template and use call method to render
- Rename arguments for consistency with design system
- Update spec to use with syntax


peteryates (approved):
🙌🏽

PR #172, opened 2021-05-27
Version 2.0.0 PanelComponent

cpjmcquillan:
- Use call method for rendering panel component
- Update spec to use with syntax


peteryates (approved):
Looks great!

PR #175, opened 2021-05-30
Version 2.0.0 - HeaderComponent

peteryates:
This is a WIP in migrating the header to the new format. The functionality is going to remain unchanged in this PR but if we want to tweak anything now's the time (cc @paulrobertlloyd)


app/components/govuk_component/header_component.html.erb
```diff
@@ -1,50 +1,45 @@
 <%= tag.header(class: classes, role: 'banner', data: { module: 'govuk-header' }, **html_attributes) do %>
   <div class="govuk-header__container govuk-width-container">
```
paulrobertlloyd, line 2: The design system component allows setting `containerClasses`; be good if `govuk-components` had a `container_classes` param, the value of which would be appended to the classes provided to this element.

```suggestion
  <div class="govuk-header__container govuk-width-container">
```

app/components/govuk_component/header_component.html.erb
```diff
@@ -1,50 +1,45 @@
...
             <image src="/assets/images/govuk-logotype-crown.png" xlink:href="" class="govuk-header__logotype-crown-fallback-image" width="36" height="32"></image>
           </svg>
```
paulrobertlloyd, line 9: Not in the design system currently, but seen often enough in the wild, it would be great if this markup could be overridden with `block` content, or be set to `nil` or similar to remove it.

See the [MoJ Design System](https://moj-design-system.herokuapp.com), [HMRC Design resources](https://design.tax.service.gov.uk) headers as well as [this comment in the community backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/97#issuecomment-552755688) as just a few examples of this need seen in the wild.

app/components/govuk_component/header_component.html.erb
```diff
@@ -1,50 +1,45 @@
...
-          <span class="govuk-header__logotype-text">
-            <%= @logo %>
-          </span>
+          <%= tag.span(@logo, class: "govuk-header__logotype-text") %>
```
paulrobertlloyd, line 10: While addressing the above, I’m not sure `logo` is the right name for this value. Maybe it should be two values: `crown` (for updating/removing the crown), and `logotype` for the text?

Or hows about nested values (showing default values):

```
logo: {
  crown: true,
  text: 'GOV.UK',
}
```

app/components/govuk_component/header_component.html.erb
```diff
@@ -1,50 +1,45 @@
 <%= tag.header(class: classes, role: 'banner', data: { module: 'govuk-header' }, **html_attributes) do %>
   <div class="govuk-header__container govuk-width-container">
     <div class="govuk-header__logo">
-      <a href=<%= @logo_href %> class="govuk-header__link govuk-header__link--homepage">
+      <%= link_to(@logo_href, class: %w(govuk-header__link govuk-header__link--homepage)) do %>
```
paulrobertlloyd, line 4: To bring parity with the design system, and prefer naming this param `homepage_url` (or if `href` is preferred, `homepage_href`).

paulrobertlloyd: Or, if using the nested params suggestion below, maybe:

```
logo: {
  crown: true,
  text: 'GOV.UK',
  href: 'https://gov.uk',
}

paulrobertlloyd:
Added a few suggestions, mainly around adapting the area around the logo, and what we’d call any new params.

Will there be a pre-release of v2.0.0? I can then update all the service where I’ve implemented these components and then if anything breaks.

paulrobertlloyd:
Another thought; rather than a separate `product_description` slot, might it be possible to just provide `product_name` as either a string or via a slot (or `block`)? 

`product_description` was added really to serve a narrow use case, chiefly that of allowing other content to be shown next to a product name such as a tag. `product_description` gives the impression this is something supported by the header design, which it is not.

peteryates:
This makes a lot of sense, I'd like to keep the number of params that the header component takes to a minimum `product_name` being a string or block seems like the best approach. Would you mind popping it in as a separate ticket and I'll address that and the other suggestions in a follow-up to this PR?

paulrobertlloyd:
Done! #177

peteryates:
Sorry I missed your question above @paulrobertlloyd. Yeah, my plan is to get all the components updated to the new slots API and then release a beta, then go through and do any tidyings up, refactoring and renaming on a second sweep, followed by a final beta/rc.

app/components/govuk_component/header_component.html.erb
peteryates: ✔️ 0ffbaf0

app/components/govuk_component/header_component.html.erb
peteryates: I've added this using a slot called `custom_logo`, the block passed in [will totally overwrite the logo contents](https://github.com/DFE-Digital/govuk-components/commit/b001e6850bc7d87bb08bbd533815c429b411d0d4#diff-6330ce0eb0a3fdda46aca3c7ecc6c519790cf8aaa4709343e363a31bcd0c1de8R75).

app/components/govuk_component/header_component.html.erb
peteryates: I opted for the two value solution, `logotype` and `crown`. Looks and feels more sensible than it did before 😃

app/components/govuk_component/header_component.html.erb
peteryates: Went for `homepage_url` in the end, it makes sense and matches the Nunjucks.

Issue #66, opened 2020-10-08
'govuk_link_to': govuk-link class not applied correctly when path parameter is a hash


benilovj:
This is a valid invocation of `link_to`:

```ruby
<%= link_to 'abc', param1: 'a', param2: 'b' %>
```
which generates:
```
<a href="/?param1=a&param2=b">abc</a>
```
---

However, calling
```ruby
<%= govuk_link_to 'abc', param1: 'a', param2: 'b' %>
```
unexpectedly generates 
```
<a href="/?class=govuk-link&param1=a&param2=b">abc</a>
```
instead of
```
<a class="govuk-class" href="/?param1=a&param2=b">abc</a>
```

(arose in https://github.com/DFE-Digital/get-help-with-tech/pull/611)

peteryates:
Strange. I'd expect that for `govuk_link_to 'abc', { param1: 'a', param2: 'b' }` the hash would end up in `args` in the [govuk_link_to](https://github.com/DFE-Digital/govuk-components/blob/master/app/helpers/govuk_link_helper.rb) helper but they end up in kwargs, so are merged with the default `govuk-link` class.

Any thoughts on a better approach that still allows custom classes to be added to the default one?

benilovj:
> Any thoughts on a better approach that still allows custom classes to be added to the default one?

Mmm, no, apart from something very hacky like an explicit if statement checking for a hash...

@tijmenb any ideas?

peteryates:
For what it's worth, I just retested what I did earlier using the dummy app rather than specs and this gives you the result you're after:

```erb
<%= govuk_link_to 'abc', { param1: 'a', param2: 'b' } %>
```

Not wrapping the query params in braces makes them end up in `kwargs` rather than `args`.

peteryates:
@benilovj - is providing the extra args in a hash a suitable fix/workaround here?

peteryates:
I'm closing this for the time being, I think the answer above covers why the behaviour occurs and how to avoid it. Agree it's not ideal, happy to revisit if it proves to remain a problem.

PR #196, opened 2021-06-20
Replace link helpers with link class helpers

### LINKED ISSUE #193: `govuk_link_to` misbehaves when given url_options without any html_options.
When `govuk_link_to` is called with url_options hash (https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to) it appends class to url_options instead of html_options, resulting in link without `govuk-link` class and with href with appended `?class=govuk-link`.

```
    <%= govuk_link_to "show", action: :show %>  #=> <a href="{correct_url}?class=govuk-link">
```
This can be avoided by adding `button: false` and parenthesising url_options:

```
    <%= govuk_link_to "show", { action: :show }, button: false %>  #=> <a class="govuk-link" href="{correct_url}">
```

Parenthesising alone does not fix the issue - url_options are then still treated as kwargs rather than part of *args (which I find a bit odd - this might depend on ruby version though)

peteryates:
Wrapping Rails' `#link_to`, `#mail_to` and `#button_to` helpers is difficult because of the internal implementation and the complications added by mixing keyword arguments in with the existing args. This difficulty is exacerbated by the change in [Ruby 3.0's separation of positional and keyword args](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/).

The above problems have been highlighted a couple of times (#66, #193) and have probably bitten other teams/developers more often.

This approach no longer attempts to wrap the methods directly but instead provides a easier way of building the GOV.UK-specific classes.

Instead of:

```slim
= govuk_link_to "some page", some_path, muted: true
```

We would now do:

```slim
= link_to "some page", some_path, class: govuk_link_classes(muted: true)
```

The `#govuk_link_classes` helper supports all the CSS options from govuk-frontend, including `:no_visited_state`, `:muted`, `:text_colour`, `:inverse`, and `:no_underline`.

The `#govuk_button_classes` helper supports `:warning`, `:disabled` and `:secondary` button styles.

Benefits of this approach:

* no complex code that juggles with arguments, optional arguments and different keyword arg handling across various versions of Ruby
* I actually understand it
* we can easily apply the styles to Rails' other link helpers, like #link_to_if, #link_to_unless and friends.

Fixes #193

### DIFF
--- a/app/components/govuk_component/summary_list_component.rb
+++ b/app/components/govuk_component/summary_list_component.rb
@@ -46,9 +46,11 @@ private
     end
 
     def action
+      link_classes = govuk_link_classes.append(action_classes).flatten
+
       tag.dd(class: "govuk-summary-list__actions") do
         if href.present?
-          govuk_link_to(href, class: action_classes, **action_attributes) do
+          link_to(href, class: link_classes, **action_attributes) do
             safe_join([text, tag.span(visually_hidden_text, class: "govuk-visually-hidden")])
           end
         end
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -1,32 +1,18 @@
 module GovukLinkHelper
-  def govuk_link_classes(no_visited_state: false, muted: false, text_colour: false, inverse: false, no_underline: false)
-    [
-      'govuk-link',
-      no_visited_state_class(no_visited_state),
-      muted_class(muted),
-      text_colour_class(text_colour),
-      inverse_class(inverse),
-      no_underline_class(no_underline),
-    ].compact
-  end
-
-  EXTRA_OPTIONS = {
-    button: false,
-    no_visited_state: false,
-    muted: false,
-    text_colour: false,
-    inverse: false,
-    no_underline: false
+  LINK_STYLES = {
+    muted: "govuk-link--muted",
+    no_visited_state: "govuk-link--no-visited-state",
+    text_colour: "govuk-link--text-colour",
+    inverse: "govuk-link--inverse",
+    no_underline: "govuk-link--no-underline",
   }.freeze
 
-  def govuk_link_to(name = nil, options = nil, extra_options = {}, &block)
+  def govuk_link_to(name = nil, options = nil, extra_options = [], &block)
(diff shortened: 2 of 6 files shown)


frankieroberto:
Like this idea!

Rather than a bunch of boolean arguments, would it make sense for `govuk_link_classes()` to accept an array of symbols instead? ie

```ruby
govuk_link_classes(:muted, :no_visited_state)
```

That way we could also allow you to add on your own classes as strings:

```ruby
govuk_link_classes(:muted, :no_visited_state, "app-link--super-important")
```

to avoid having to do this:

```ruby
link_to "blah", "/", class: govuk_link_classes(:muted, :no_visited_state).push("app-link--super-important")
```

edwardhorsford:
I might be wary of this - as I think it makes it more likely that developers would forget it.

it's quite common for developers to attempt to style things as buttons or links using the classes alone - my hunch would be that having a named `govuk` thing to use helps reinforce that most things should be using `govuk` versions rather than native.

cpjmcquillan:
I'd also throw in my 2p that... 
```ruby
link_to "something", "#here", class: govuk_link_classes(:some_option)
```

Doesn't feel much different to...
```ruby
link_to "something", "#here", class: "all-the classes-i want-to add govuk-link"
```

I'd be wary of the few instances where the existing helper isn't suitable driving a fairly significant change to something that is really useful to lots of teams in the vast majority of use cases!

peteryates:
Thanks for the feedback all.

@BroiSatse has made an excellent attempt at implementing a version of `#govuk_link_to` that does the necessary argument juggling (#197) and I think the two can live side-by-side, ideally with the link helpers using the class helpers.

Going to see if I can combine the to approaches into one PR. Also using symbols seems like a good idea, will look into that too 💡

peteryates:
Ok, so this PR took a little detour but now we have:

* new link and button class helpers that can be injected into any of Rails myriad of link helpers
* re-written wrappers for `#link_to`, `#button_to` and `#mail_to` that properly add the GOV.UK classes without (hopefully, at least) affecting the underlying Rails functionality. Many thanks to @BroiSatse for his help on this

PR #197, opened 2021-06-21
Links with url params

### LINKED ISSUE #193: `govuk_link_to` misbehaves when given url_options without any html_options.

BroiSatse:
Please don't judge me, this is pretty ugly - but works!

It is also potentially possible to get it working with kwargs, but would require special handling when kwargs caputres url_options - so it would still require defining all the arguments and special handling in case we have been provided block. Honestly `link_to` signature is... amazing...

This fixes #193 and #66.

### DIFF
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -1,8 +1,27 @@
 module GovukLinkHelper
-  def govuk_link_to(*args, button: false, no_visited_state: false, muted: false, text_colour: false, inverse: false, no_underline: false, **kwargs, &block)
-    classes = build_classes(button, no_visited_state, muted, text_colour, inverse, no_underline)
+  EXTRA_OPTIONS = {
+    button: false,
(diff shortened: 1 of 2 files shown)


peteryates:
Closing this PR as I folded your changes into #196

PR #231, opened 2021-08-04
Don't error when summary list value is empty

peteryates:
Some services might want to display the row in the summary list despite it having no content.


spec/components/govuk_component/summary_list_component_spec.rb
```diff
@@ -218,8 +218,10 @@
   it_behaves_like 'a component that accepts custom HTML attributes'
 
   context "when there is no text or block" do
+    subject! { render_inline(described_class.new) }
+
     specify "raises an appropriate error" do
```
cpjmcquillan, line 223: "renders an empty row"

paulrobertlloyd (approved):
LGTM!

spec/components/govuk_component/summary_list_component_spec.rb
peteryates: Ah damn, thanks @cpjmcquillan!

PR #243, opened 2021-08-21
Tables

### LINKED ISSUE #53: Table component

peteryates:
This is going to be the most complex component supported by the library. After some experimentation it's likely that not all the Nunjucks options will be supported out-of-the-box, mainly due to the inability for slots to access their parents.

The entire set of functionality (and more) will be available when building via slots, however.

### With a multidimensional array

This approach is most like that provided by the Nunjucks macros. It will be fine for most use cases but doesn't offer much in the way of flexibility. We can follow the Nunjucks arguments closely here.

```erb
<%= govuk_table(head: %w(c1 c2 c3), rows: [%w(r1c1 r1c2 r1c3), %w(r2c1 r2c2 r2c3)], first_cell_is_header: true) %>
```

### With slots

Clearly more code is required here but every `table`, `head`, `body`, `row`, and `cell` will be a fully-fledged `GovukComponent::Base` and, as a result can be passed any custom CSS classes, HTML attributes or block content. This should allow for complex layouts with column and row spans, if they're needed.

```erb
<%= govuk_table do |table| %>
  <% table.caption('A wild table appears', size: 'l') %>
  <% table.head do |head| %>
    <% head.row do |row| %>
      <% row.cell(header: true, text: 'c1') %>
      <% row.cell(header: true, text: 'c2', numeric: true) %>
      <% row.cell(header: true, text: 'c3', width: 'one-half') %>
    <% end %>
  <% end %>

  <% table.body do |body| %>
    <% body.row do |row| %>
      <% row.cell(text: 'r1c1') %>
      <% row.cell(text: 'r1c2', numeric: true) %>
      <% row.cell(text: 'r1c3') %>
    <% end %>

    <% body.row do |row| %>
      <% row.cell(text: 'r2c1') %>
      <% row.cell(text: 'r2c2', numeric: true) %>
      <% row.cell(classes: 'with-arbitrary-block-content') do %>
        r2c3
      <% end %>
    <% end %>
  <% end %>
<% end %>
```

Additionally, [captions](https://design-system.service.gov.uk/components/table/#table-captions) will be supported as slots and unlike the original macro take a `size:` argument that can be `s`, `m`, `l` or `xl` in the same format as the [form builder's captions](https://govuk-form-builder.netlify.app/introduction/labels-captions-hints-and-legends/#a-label-with-a-caption).

## Tasks

* [x] Allow for multiple `tbody` elements (I _think_ this is just a case of changing `renders_one :tbody` to `renders_many :tbodys` but Rails' inflector might want `renders_many :tbodies`!)
* [x] Add custom HTML support in the caption
* [x] See how well having a `caption` argument on the `TableComponent` works, might make sense to follow the Nunjucks API
* [x] Support `first_cell_is_header` to add headers per row
* [x] Add options for custom width arguments

Fixes #53


frankieroberto:
I wonder if the component should require a caption (for accessibility)?

That said though, I think I would be tempted to just use plain ERb for tables, as they can be quite complicated, and the benefits of being a component/macro seem less than for other components.

peteryates:
Yeah, good call regarding the caption being a requirement. Currently it's not in the Nunjucks counterpart so making it required might need to be added as a setting rather than enforced everywhere.

I'm mostly with you on just using ERb for tables too, this is really a case of plugging the final gap in the library. I think the ability to generate a whole table from an array of rows will be a benefit for some, but by the time you've used the DSL-style builder (like in the 'with slots' example) you've nearly got as many lines as you would have in ERb anyway 😓

frankieroberto:
The nunchucks macro lets you specify a caption like this:

```
{{ govukTable({
  caption: "Dates and amounts",
  captionClasses: "govuk-table__caption--m",
  ...
)} }}
```

(Intriguingly, this doesn't follow the `html` and `text` pattern, although I putting HTML in the caption would be quite rare.)

frankieroberto:
Oh, the other thing to mention is that a table can have multiple `<tbody>` elements – useful for grouping rows, which can then be associated with `<th scope="rowgroup">`. I've actually done this in one service...

Oh, and there's also a `<tbody>` element that we _might_ want to support, for completeness? (Although the nunjucks macro doesn't)

peteryates:
So in this early version of the component you'd do the above like this:

```
<%= govuk_table do |table| %>
  <%= table.caption(text: "Dates and amounts", size: "m") %>
<% end %>
```

Or you'll be able to do the same with a block if you want custom HTML (not added yet but it'll be there before too long):

```
<%= govuk_table do |table| %>
  <%= table.caption(size: "l", classes: %w[some-other-classes]) do %>
    <span>Dates and amounts</span>
  <% end %>
<% end %>
```

It would be quite straightforward to allow the caption to be passed in as an arg too, but to me it feel like the slot's a good fit here.

PR #244, opened 2021-08-29
Refresh the documentation

peteryates:
This replaces the current [documentation page](https://dfe-digital.github.io/govuk-components/) with one built with [Lookbook](https://github.com/allmarkedup/lookbook).

The approach has some advantages:

* Nice UI for browsing variations
* Has built in resizing tool for previewing responsiveness
* Supports searching/filtering
* Won't require manually refreshing the examples page
* Uses view-component's native [preview](https://viewcomponent.org/guide/previews.html) functionality

## Things to solve

* [ ] hosting - we'll need to host the dummy Rails app somewhere, GOV.UK PaaS makes sense
* [x] formatting the output - currently it's all on one line, would be nice to make it more readable (see allmarkedup/lookbook/issues/39)
* [ ] remove examples that just pass a string in via a block, they give the impression you can pass in any object but that's not the case, see #249 
* [x] work out how we'll document helper methods (might even mean choosing something other than Lookbook 😞 )

## Preview

Screenshot-2021-08-29-12:08:18


paulrobertlloyd:
Oooh, lovely stuff. How can I get this running locally?

peteryates:
It's set up in the 'dummy' Rails app in `spec/dummy`.

Just bundle as normal, then `cd spec/dummy` and `bundle exec rails s`.

It's the root route, so the app should then be available at localhost:3000.

peteryates:
Screenshot-2021-09-04-14:41:44

Coming together quite nicely!

paulrobertlloyd:
Ah, just needed the right combination of commands to get the WEBrick server to start. This should be really helpful to those using the gem. I wonder if the examples should align with those on the design system website, where possible?

paulrobertlloyd:
Oh hell yeah! Been waiting for this. Guessing the docs will be hosted on a new URL?

peteryates:
Yeah, planning on getting it sorted over the Christmas break - just need to work out how to make it automatically deploy to the PaaS :)

PR #253, opened 2021-09-23
Allow links with different HTTP method in header

cesidio:
In order to logout, Devise uses HTTP delete endpoints, so if you want to have a nav item to logout, this allows you to do that

### DIFF
--- a/app/components/govuk_component/header_component.rb
+++ b/app/components/govuk_component/header_component.rb
@@ -53,14 +53,15 @@ private
   end
 
   class NavigationItem < GovukComponent::Base
-    attr_reader :text, :href, :active
+    attr_reader :text, :href, :options, :active
 
-    def initialize(text:, href: nil, active: false, classes: [], html_attributes: {})
+    def initialize(text:, href: nil, options: {}, active: false, classes: [], html_attributes: {})
       super(classes: classes, html_attributes: html_attributes)
 
-      @text   = text
-      @href   = href
-      @active = active
+      @text    = text
+      @href    = href
+      @options = options
+      @active  = active
     end
 
     def active?
@@ -77,7 +78,7 @@ private
 
     def call
       if link?
-        link_to(text, href, class: "govuk-header__link")
+        link_to(text, href, **options, class: "govuk-header__link")
       else
         text
       end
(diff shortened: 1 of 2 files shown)


david-mears-dfe (approved):
LGTM

frankieroberto:
Would it be worth adding an optional `options` param, and passing that to the Rails `link_to` helper instead? That way `remote: true` would also be supported, as well as any other options that `link_to` might support.

app/components/govuk_component/header_component.rb
```diff
@@ -77,7 +78,12 @@ def link?
 
     def call
       if link?
-        link_to(text, href, class: "govuk-header__link")
+        case method
+        when :get
+          link_to(text, href, class: "govuk-header__link")
+        when :post, :patch, :put, :delete
+          link_to(text, href, method: method, class: "govuk-header__link")
+        end
```
cpjmcquillan, line 86: Should we use `button_to` here instead?

I'm not sure how that would affect styling though.

cpjmcquillan (approved):
@cesidio I miss you 😭

🚀

app/components/govuk_component/header_component.rb
cesidio: Yes, that would affect the styling as we have links on navbar

cpjmcquillan: My concern is that links with other actions won't progressively enhance. I wonder if it's possible to style the button as a navigation item link.

cesidio:
> Would it be worth adding an optional `options` param, and passing that to the Rails `link_to` helper instead? That way `remote: true` would also be supported, as well as any other options that `link_to` might support.

It makes sense what you said, I have done that. Can you please check if it makes more sense now?

frankieroberto:
> > Would it be worth adding an optional `options` param, and passing that to the Rails `link_to` helper instead? That way `remote: true` would also be supported, as well as any other options that `link_to` might support.
> 
> It makes sense what you said, I have done that. Can you please check if it makes more sense now?

Perfect! 👌

frankieroberto (approved):

peteryates (approved):
Looks great to me.

PR #254, opened 2021-09-23
Automatic active nav item on current page

cesidio:
This is more for a discussion, I guess that if you are on the current_page you want the nav item active, cannot think of cases when you do not want that.

### DIFF
--- a/spec/components/govuk_component/header_component_spec.rb
+++ b/spec/components/govuk_component/header_component_spec.rb
@@ -1,6 +1,13 @@
 require 'spec_helper'
 
 RSpec.describe(GovukComponent::HeaderComponent, type: :component) do
+  before do
+    allow_any_instance_of(GovukComponent::HeaderComponent::NavigationItem).to(
+      receive(:request).and_return(double(ActionDispatch::Request, get?: true, path: current_page))
+    )
+  end
+
+  let(:current_page) { "/item-3" }
   let(:component_css_class) { 'govuk-header' }
 
   let(:product_name) { 'Order an amazing ID' }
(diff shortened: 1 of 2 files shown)


peteryates:
@cesidio - I think this is failing because the `current_page?` helper needs to be called in [`before_render`](https://viewcomponent.org/guide/lifecycle.html#before_render) rather than on initialize.

cesidio:
> @cesidio - I think this is failing because the `current_page?` helper needs to be called in [`before_render`](https://viewcomponent.org/guide/lifecycle.html#before_render) rather than on initialize.

@peteryates I tested it and it works fine, it's just the spec that is broken.

peteryates:
@cacquani ah I see, I just pushed a tweak that stubs the appropriate request on the NavigationItem. Feel free to force push over if you think there's a better way!

spec/components/govuk_component/header_component_spec.rb
```diff
@@ -1,6 +1,12 @@
 require 'spec_helper'
 
 RSpec.describe(GovukComponent::HeaderComponent, type: :component) do
+  before do
+    allow_any_instance_of(GovukComponent::HeaderComponent::NavigationItem).to(
+      receive(:request).and_return(double(ActionDispatch::Request, get?: true, path: "/a-path"))
+    )
+  end
```
cpjmcquillan, line 8: Can we use something like this from the [docs](https://viewcomponent.org/guide/testing.html#setting-requestpath_parameters)?

```ruby
context "when the link is the current page" do
  it "renders an active navigation item" do
    with_request_url "/a-path" do
      component = Component.new ...

      render_inline(component)

      expect(rendered_component).to have_css "active navigation item"
    end
  end
end
```

peteryates: Missed that! That's way nicer - feel free to replace my commit.

cesidio: We tried this and it led us down to a horrible route :(

We need to add more dependencies to the gem because `with_request_url` helper requires more helpers that we do not include in the gem.

peteryates: I'm happy with this approach. I had a quick go on Friday and couldn't get it working nicely. The stub approach isn't _too bad_, it's quite clear what's going on.

peteryates (approved):
This looks fine to me, seems like a sensible thing to default 👍🏽

cpjmcquillan:
Just checking, but does this definitely work? I'm surprised we don't need to delegate `current_page?` to `helpers`

cesidio:
> Just checking, but does this definitely work? I'm surprised we don't need to delegate `current_page?` to `helpers`

I just double checked and yes it works! Not sure why we do not need to delegate `current_page?` to `helpers`

cpjmcquillan (approved):

Issue #256, opened 2021-10-05
Reference to fallback crown image is hardcoded


frankieroberto:
The header component contains a fallback `<img>` tag, for when a browser doesn’t support inline SVG.

The `src` of the tag [is hardcoded](https://github.com/DFE-Digital/govuk-components/blob/master/app/components/govuk_component/header_component.html.erb#L12) to `/assets/images/govuk-logotype-crown.png`.

However there is no guarantee that the image will be available at this URL, depending on how the asset pipeline, etc is set up – and the image URL currently 404s on many of our service, for example:

* https://www.apply-for-teacher-training.service.gov.uk/assets/images/govuk-logotype-crown.png
* https://www.find-postgraduate-teacher-training.service.gov.uk/assets/images/govuk-logotype-crown.png
* https://www.register-trainee-teachers.education.gov.uk/assets/images/govuk-logotype-crown.png

As a short-term fix, we could replace the hardcode url with a param, and update our services so that they pass in a path to that image?

peteryates:
Yeah, that sounds like a good solution to me. I guess the longer term fix would be to allow these things to be set via config, but having a path override will make sense either way.

frankieroberto:
Yeah, the trickier problem is somehow getting services to remember to do this, as it'd require a bit of manual work to copy the right file to the right folder and then figure out how to reference it (depending on how webpacker, etc is set up). I think I'd be tempted to put it in the `/public` folder and skip the asset pipeline completely.

If the path isn't specified, should we not include the image at all, or raise an exception to force it to be specified (would be a breaking change)?

peteryates:
Fixed in #257

PR #257, opened 2021-10-05
Allow crown's fallback image to have custom path

peteryates:
The previously-hardcoded value makes a lot of assumptions on how the application is bundling its assets, this should offer a little more flexibility.

Refs #256


app/components/govuk_component/header_component.rb
```diff
@@ -52,6 +55,16 @@ def container_classes
     combine_classes(%w(govuk-header__container govuk-width-container), custom_container_classes)
   end
 
+  def crown_fallback_image_attributes
+    {
+      src: crown_fallback_image,
+      class: "govuk-header__logotype-crown-fallback-image",
+      width: "36",
+      height: "32",
+      "xlink:href" => "",
```
peteryates, line 64: This [appears to be a deprecated attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href) and it's blank - safe to drop it?

frankieroberto: I _think_ this was deliberately included, as a work-around for some legacy browser bug.

But looking at the current HTML for [the header](https://design-system.service.gov.uk/components/header/) in the Design System, it looks like it has been dropped, and the image has been wrapped in IE conditional comments now:

```html
      <a href="#" class="govuk-header__link govuk-header__link--homepage">
        <span class="govuk-header__logotype">
          <!--[if gt IE 8]><!-->
          <svg aria-hidden="true" focusable="false" class="govuk-header__logotype-crown" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 132 97" height="30" width="36">
            <path fill="currentColor" fill-rule="evenodd" d="M25 30.2c3.5 1.5 7.7-.2 9.1-3.7 1.5-3.6-.2-7.8-3.9-9.2-3.6-1.4-7.6.3-9.1 3.9-1.4 3.5.3 7.5 3.9 9zM9 39.5c3.6 1.5 7.8-.2 9.2-3.7 1.5-3.6-.2-7.8-3.9-9.1-3.6-1.5-7.6.2-9.1 3.8-1.4 3.5.3 7.5 3.8 9zM4.4 57.2c3.5 1.5 7.7-.2 9.1-3.8 1.5-3.6-.2-7.7-3.9-9.1-3.5-1.5-7.6.3-9.1 3.8-1.4 3.5.3 7.6 3.9 9.1zm38.3-21.4c3.5 1.5 7.7-.2 9.1-3.8 1.5-3.6-.2-7.7-3.9-9.1-3.6-1.5-7.6.3-9.1 3.8-1.3 3.6.4 7.7 3.9 9.1zm64.4-5.6c-3.6 1.5-7.8-.2-9.1-3.7-1.5-3.6.2-7.8 3.8-9.2 3.6-1.4 7.7.3 9.2 3.9 1.3 3.5-.4 7.5-3.9 9zm15.9 9.3c-3.6 1.5-7.7-.2-9.1-3.7-1.5-3.6.2-7.8 3.7-9.1 3.6-1.5 7.7.2 9.2 3.8 1.5 3.5-.3 7.5-3.8 9zm4.7 17.7c-3.6 1.5-7.8-.2-9.2-3.8-1.5-3.6.2-7.7 3.9-9.1 3.6-1.5 7.7.3 9.2 3.8 1.3 3.5-.4 7.6-3.9 9.1zM89.3 35.8c-3.6 1.5-7.8-.2-9.2-3.8-1.4-3.6.2-7.7 3.9-9.1 3.6-1.5 7.7.3 9.2 3.8 1.4 3.6-.3 7.7-3.9 9.1zM69.7 17.7l8.9 4.7V9.3l-8.9 2.8c-.2-.3-.5-.6-.9-.9L72.4 0H59.6l3.5 11.2c-.3.3-.6.5-.9.9l-8.8-2.8v13.1l8.8-4.7c.3.3.6.7.9.9l-5 15.4v.1c-.2.8-.4 1.6-.4 2.4 0 4.1 3.1 7.5 7 8.1h.2c.3 0 .7.1 1 .1.4 0 .7 0 1-.1h.2c4-.6 7.1-4.1 7.1-8.1 0-.8-.1-1.7-.4-2.4V34l-5.1-15.4c.4-.2.7-.6 1-.9zM66 92.8c16.9 0 32.8 1.1 47.1 3.2 4-16.9 8.9-26.7 14-33.5l-9.6-3.4c1 4.9 1.1 7.2 0 10.2-1.5-1.4-3-4.3-4.2-8.7L108.6 76c2.8-2 5-3.2 7.5-3.3-4.4 9.4-10 11.9-13.6 11.2-4.3-.8-6.3-4.6-5.6-7.9 1-4.7 5.7-5.9 8-.5 4.3-8.7-3-11.4-7.6-8.8 7.1-7.2 7.9-13.5 2.1-21.1-8 6.1-8.1 12.3-4.5 20.8-4.7-5.4-12.1-2.5-9.5 6.2 3.4-5.2 7.9-2 7.2 3.1-.6 4.3-6.4 7.8-13.5 7.2-10.3-.9-10.9-8-11.2-13.8 2.5-.5 7.1 1.8 11 7.3L80.2 60c-4.1 4.4-8 5.3-12.3 5.4 1.4-4.4 8-11.6 8-11.6H55.5s6.4 7.2 7.9 11.6c-4.2-.1-8-1-12.3-5.4l1.4 16.4c3.9-5.5 8.5-7.7 10.9-7.3-.3 5.8-.9 12.8-11.1 13.8-7.2.6-12.9-2.9-13.5-7.2-.7-5 3.8-8.3 7.1-3.1 2.7-8.7-4.6-11.6-9.4-6.2 3.7-8.5 3.6-14.7-4.6-20.8-5.8 7.6-5 13.9 2.2 21.1-4.7-2.6-11.9.1-7.7 8.8 2.3-5.5 7.1-4.2 8.1.5.7 3.3-1.3 7.1-5.7 7.9-3.5.7-9-1.8-13.5-11.2 2.5.1 4.7 1.3 7.5 3.3l-4.7-15.4c-1.2 4.4-2.7 7.2-4.3 8.7-1.1-3-.9-5.3 0-10.2l-9.5 3.4c5 6.9 9.9 16.7 14 33.5 14.8-2.1 30.8-3.2 47.7-3.2z"></path>
          </svg>
          <!--<![endif]-->
          <!--[if IE 8]>
          <img src="/assets/images/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image" width="36" height="32">
          <![endif]-->
          <span class="govuk-header__logotype-text">
            GOV.UK
          </span>
        </span>
      </a>
```

frankieroberto: Ah yes, this was changed back in May: [Change approach to fallback PNG in the header to fix blank data URI from triggering CSP error](https://github.com/alphagov/govuk-frontend/pull/2229)

frankieroberto:
I wonder if it'd be safest, for now, to not include the image at all if the URL isn't given? (As the fallback URL is quite likely to be wrong)

peteryates:
Yeah - I agree. I've pushed another commit that removes it unless a `crown_fallback_image` is provided.

app/components/govuk_component/header_component.rb
```diff
@@ -52,6 +55,15 @@ def container_classes
     combine_classes(%w(govuk-header__container govuk-width-container), custom_container_classes)
   end
 
+  def crown_fallback_image_attributes
+    {
+      class: "govuk-header__logotype-crown-fallback-image",
+      width: "36",
+      height: "32",
+      "xlink:href" => "",
```
frankieroberto, line 63: Looks like we can drop this now that we're using `` with IE8 conditional comments.

```suggestion
```

frankieroberto (changes requested):
Looks good. I think we just need to drop the `xlink:href:` attribute now.

Also maybe `crown_fallback_image` should be named `crown_fallback_image_path` just to be super explicit that we’re expecting a path to be set here?

frankieroberto (approved):

PR #262, opened 2021-10-20
Hide the cookie banner by default when printing

### LINKED ISSUE #261: Hide the cookie banner from printouts by default
This feels like the right move to me; can't imagine any cases where people printing out pages would want a cookie banner at the top!

Something along the lines of:

```
def default_classes
  %w(govuk-cookie-banner).tap do |c|
    c << 'govuk-!-display-none-print' if disable_when_printing?
  end
end
```

See DFE-Digital/apply-for-teacher-training/pull/5882

cpjmcquillan:
Before, the cookie banner component was always displayed.

It makes sense to add the `govuk-!-display-none-print` class by
default so that the cookie banner is hidden when printing a
page.

This can be overriden by setting `hide_when_printing` to false.

Closes #261

### DIFF
--- a/app/components/govuk_component/cookie_banner_component.rb
+++ b/app/components/govuk_component/cookie_banner_component.rb
@@ -1,13 +1,14 @@
 class GovukComponent::CookieBannerComponent < GovukComponent::Base
   renders_many :messages, GovukComponent::CookieBannerComponent::MessageComponent
 
-  attr_accessor :aria_label, :hidden
+  attr_accessor :aria_label, :hidden, :hide_in_print
 
-  def initialize(aria_label: "Cookie banner", hidden: false, classes: [], html_attributes: {})
+  def initialize(aria_label: "Cookie banner", hidden: false, hide_in_print: true, classes: [], html_attributes: {})
     super(classes: classes, html_attributes: html_attributes)
 
-    @aria_label = aria_label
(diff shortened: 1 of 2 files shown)


app/components/govuk_component/cookie_banner_component.rb
```diff
@@ -1,13 +1,14 @@
 class GovukComponent::CookieBannerComponent < GovukComponent::Base
   renders_many :messages, GovukComponent::CookieBannerComponent::MessageComponent
 
-  attr_accessor :aria_label, :hidden
+  attr_accessor :aria_label, :hidden, :hide_when_printing
 
-  def initialize(aria_label: "Cookie banner", hidden: false, classes: [], html_attributes: {})
+  def initialize(aria_label: "Cookie banner", hidden: false, hide_when_printing: true, classes: [], html_attributes: {})
```
cpjmcquillan, line 6: I'm not convinced `hide_when_printing` is the nicest variable name here. 

I went with `hide_when_printing: true` over `display_when_printing: false` because it made assigning the class using `class_names` more readable.

peteryates: I just checked and there's a similar option in the breadcrumbs component called `hide_in_print` - I think it'd make sense for them to be consistent.

Sorry, didn't realise until now!

PR #264, opened 2021-10-20
Use class_names helper method to conditionally set classes

cpjmcquillan:
Before, classes were conditionally added by tapping an array of classes and
conditionally appending classes.

Now, the `class_names` helper is used consistently to add conditional classes.


cpjmcquillan:
Throwing this PR out there as a suggestion.

I think the `class_names` [method](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-token_list) is a nice way of conditionally adding classes. 

If we prefer to keep things the way they are, then that is ok too.

PR #266, opened 2021-12-01
add semicolon to copyright html entity

alexbowen:
being a bit ocd maybe but IE throws a little warning for this

Screenshot 2021-12-01 at 14 16 36

### DIFF
--- a/app/components/govuk_component/footer_component.rb
+++ b/app/components/govuk_component/footer_component.rb
@@ -77,7 +77,7 @@ private
   end
 
   def default_copright_text
-    raw(%(&copy Crown copyright))
+    raw(%(&copy; Crown copyright))
   end
 
   def default_copyright_url


app/components/govuk_component/footer_component.rb
```diff
@@ -77,7 +77,7 @@ def build_copyright(text, url)
   end
 
   def default_copright_text
-    raw(%(&copy Crown copyright))
+    raw(%(&copy; Crown copyright))
```
frankieroberto, line 80: Encode it directly?

```suggestion
    raw(%(© Crown copyright))
```

peteryates (approved):
Looks good. Are there any disadvantages in the direct encoding?

frankieroberto:
> Looks good. Are there any disadvantages in the direct encoding?

It might break if you used a different character encoding than UTF-8... But that should be rare nowadays I think. And I think even if you used ISO-8859-1 (which is the default on the web if you don't specify) it might still work?

PR #279, opened 2021-12-28
Guide

peteryates:
This is _another_ attempt at building a guide. The previous [Lookbook](https://github.com/allmarkedup/lookbook) is great but doesn't allow any descriptive text and needs a more complex hosting setup.

This version is more like the [form builder's](https://govuk-form-builder.netlify.app/), it builds a static Nanoc site and pulls in the actual code from the library so can't fall out of date.

## Tasks

* [x] get a Nanoc site bootstrapped and make it able to render the components
* [x] add the GOV.UK Frontend stuff and make it look like the form builder guide
* [x] write the actual guide pages
* [x] create an image for the masthead. @paulrobertlloyd - would you be able to lend a hand with this bit? Nice matching ones on the form builder and here would be amazing. You're more than welcome to choose a colour scheme/accent too if the purple's a bit much
* [x] add a page covering common options across components/helpers, mainly `classes` and `html_attributes`.

## Bugs

* [x] work out why the heading menu only works when there's a single heading on the page - probably a clashing id or something


paulrobertlloyd:
@peteryates Have I got write permissions for this repository? Want to push a commit to this branch. Cheers!

paulrobertlloyd:
🙏

peteryates:
Oops, thought you had them! Should see an invite now.

peteryates:
The examples pages are now present in draft form, they need some context and extra information and guidance but everything's looking decent.

The list on the homepage is a bit long, what are your thoughts on spreading it over two columns @paulrobertlloyd?

Something like this (just by adding `columns: 2` to the `ul`)

| Before | After |
| ------ | ------- |
| Screenshot from 2022-01-01 18-51-55 | Screenshot from 2022-01-01 18-51-42 |

peteryates:
Also there's a couple of things that might need a reorganise; in this library the skip link is technically a helper method rather than a component, but it's listed under components as that's what it is in the source Nunjucks. Perhaps moving 'Link Helpers' from the introduction to Comonents and renaming the list "Components and helpers"?

paulrobertlloyd:
Screenshot 2022-01-01 at 21 01 56

Split components into 2 columns as suggested. Have split link and button helpers out into separate pages, and placed under a ‘Helpers’ heading. Do we also need to include `govuk_mail_to` helper? Are there other helpers?

paulrobertlloyd:
What makes the skip link functionally different from a component, apart from where it’s in the codebase?

peteryates:
Nothing really other than you can't create it via `render GovukComponent::SkipLinkComponent.new`, only via `govuk_skip_libk`. Now they're separate I think moving it to the helper section makes more sense really.

paulrobertlloyd:
> Now they're separate I think moving it to the helper section makes more sense really.

I agree.

peteryates:
So there's now a first draft for each component.

The reason the header example page isn't working properly is that in `govuk-frontend/govuk/all.js` there's this:

```javascript
// Find first header module to enhance.
var $toggleButton = scope.querySelector('[data-module="govuk-header"]');
new Header($toggleButton).init();
```

If I change it to this (manually, then copy the file to the right place) it works:

```javascript
?
// Find all header modules and enhance them
var $toggleButtons = scope.querySelectorAll('[data-module="govuk-header"]');
nodeListForEach($toggleButtons, function ($toggleButton) {
  new Header($toggleButton).init();
});
```

Not really sure on the best way of doing this cleanly (where cleanly means not doing some `sed` gymnastics in the `postinstall`). Any ideas @paulrobertlloyd?

paulrobertlloyd:
I've been thinking about this. Ultimately, I think component examples should be in an `iframe` – note how the success banner example always receives focus when go to that page. Heading levels are wrong too. Using an iFrame, as they do on the Design System docs, will ensure component examples remain encapsulated.

In the short term, what happens if you set `html_attributes.data.module` to `nil` or `null` (or similar)?

paulrobertlloyd:
Possibly worth looking at:

* https://github.com/alphagov/govuk-design-system/blob/main/views/partials/_example.njk
* https://github.com/alphagov/govuk-design-system/blob/main/src/stylesheets/components/_example.scss
* https://github.com/alphagov/govuk-design-system/blob/main/src/javascripts/components/example.js

peteryates:
Yeah - I had a quick look at how the `<iframe>` resizing worked in the official guide and it put me off! As we've only got a couple here and they don't look too bad, I'll leave it for the moment and revisit.

Overall I think it looks great, the content will evolve over time but I think it's fine for the first version.

paulrobertlloyd:
Yeah, this is fine, and a perfect solution to the immediate issues.

Meanwhile… I’m seeing issues running Nanoc on Ruby 3.1.0:

```
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Listening for lib/ changes…
View the site at http://127.0.0.1:3000/
[2022-01-03 10:50:17] INFO  WEBrick 1.7.0
[2022-01-03 10:50:17] INFO  ruby 3.1.0 (2021-12-25) [x86_64-darwin21]
[2022-01-03 10:50:17] INFO  WEBrick::HTTPServer#start: pid=66124 port=3000
/Users/paulrobertlloyd/.rvm/gems/ruby-3.1.0/gems/railties-7.0.0/lib/rails/railtie.rb:246:in `initialize': Rails::Engine is abstract, you cannot instantiate it directly. (RuntimeError)
```

Is the components gem supported by Ruby 3.1.0, too?

guide/lib/helpers/content_helpers.rb
```diff
@@ -0,0 +1,128 @@
...
+        "GovukComponent::Accordion" => "govuk_accordion",
+        "GovukComponent::BackLink" => "govuk_back_link",
+        "GovukComponent::Breadcrumbs" => "govuk_breadcrumbs",
+        "GovukComponent::CookieBanner" => "govuk_cookie_banner",
+        "GovukComponent::Details" => "govuk_details",
+        "GovukComponent::Footer" => "govuk_footer",
+        "GovukComponent::Header" => "govuk_header",
+        "GovukComponent::InsetText" => "govuk_inset_text",
+        "GovukComponent::NotificationBanner" => "govuk_notification_banner",
+        "GovukComponent::Panel" => "/components/panel",
```
paulrobertlloyd, line 117: Should this be this?

```suggestion
        "GovukComponent::Panel" => "govuk_panel",
```

peteryates: 🤦🏽‍♂️

peteryates:
Ruby 3.1 isn't yet supported by Rails, we need to wait for Rails 7.0.1. The original plan for this release was to include it but it might need to follow shortly after.

PR #290, opened 2022-01-09
Add publish github action

peteryates:
Until now I've been building this on my own machine but that isn't ideal. 3.0.0 was pinned to Rails 6.1.4.4 because I had RAILS_VERSION set in my `~/.envrc`.

It's cleaner if we publish via a GitHub action.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,33 @@
+name: Publish
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Tag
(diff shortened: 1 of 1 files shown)


baarkerlounger:
Is the version tag being `3.0.1b1` rather than `3.0.1` an artefact of this or is there another reason?

peteryates:
No, was just [testing it out](https://github.com/dfe-digital/rails-template). The beta release now has the correct requirements so I'll release `3.0.1` tomorrow.

PR #316, opened 2022-05-15
Pagination

peteryates:
Add the PaginationComponent.

An official GOV.UK pagination component [is on its way](/govuk-frontend/pull/2610) so this PR is just laying the groundwork for when it's ready.

It uses [Pagy](https://github.com/ddnexus/pagy), a pagination Gem that works well with any Rack app (or plain Ruby). Pagy was picked over [Kaminari](https://github.com/kaminari/kaminari) because it's more modular, less Rails-specific, and easier to wrap a ViewComponent around.

## Remaining tasks

* [x] Create a draft guide page, refer [to the offical draft for reference](https://add-pagination--govuk-design-system-preview.netlify.app/components/pagination/) (PR [here](/alphagov/govuk-design-system/pull/1972))
* [x] Add some configuration options so the 'Next' and 'Previous' can modified, eg. 'Previous team' and 'Next team'. This'll have to overwrite the `govuk-visually-hidden` text so it should be able to take a chunk of HTML as well as just a string.
* [x] Test it properly with the real CSS once it's published (or a beta is made available)
* [x] Work out the best thing to do regarding the ellipse character. `&ctdot;` seems to get eaten by Nokogiri and makes the tests a bit tricker to write. It's using the unicode character `⋯` for the time being (just leaving the unicode character in for the time being, we've used them elsewhere without any problems)
* [x] Dont' render at all when there's just one page?
* [x] Investigate [decoration](https://github.com/alphagov/govuk-frontend/pull/2610/files#diff-8f24600b8b58d8a6ff19d90cadc27de09b8314b4405af61913e4e509858d39a4R107) (the `govuk-pagination__link-title--decorated` is added to prev/next titles when there is no `label_text` present)
* [ ] Add pagination links to the README


app/components/govuk_component/pagination_component.rb
```diff
@@ -0,0 +1,191 @@
...
+
+  attr_reader :pagy
+
+  renders_one :previous_item, ->(href:, text: "Previous", label: nil, classes: [], html_attributes: {}) do
+    PreviousItem.new(href: href, text: text, label: label, block_mode: block_mode?, classes: classes, html_attributes: html_attributes)
+  end
+
+  renders_one :next_item, ->(href:, text: "Next", label: nil, classes: [], html_attributes: {}) do
+    NextItem.new(href: href, text: text, label: label, block_mode: block_mode?, classes: classes, html_attributes: html_attributes)
+  end
```
peteryates, line 12: We need to use a [lambda slot](https://viewcomponent.org/guide/slots.html#lambda-slots) here so we can pass the `block_mode?` down to the children.

It's only _really_ needed on the `NextItem`, because in block mode the arrow goes before the title (`→ Next` instead of `Next →`), but applying it to both at least makes them match 😅

spec/components/govuk_component/pagination_component_spec.rb
```diff
@@ -0,0 +1,255 @@
+require 'spec_helper'
+
+module Pagy::UrlHelpers
+  def request
+    OpenStruct.new(GET: {}, session: nil, host: "https://somesite/")
+  end
+end
```
peteryates, line 7: I suspect there's a nicer way of doing this, the usual 'pretend it's a request' tricks in RSpec weren't playing ball.

cpjmcquillan: Have you tried the [`with_request_url` test helper](https://viewcomponent.org/guide/testing.html#setting-requestpath_parameters)?

peteryates: Yeah, that was my first port of call. It wasn't being picked up from within Pagy.

app/components/govuk_component/pagination_component.rb
```diff
@@ -0,0 +1,191 @@
...
+        end
+      end
+    end
+
+  private
+
+    def title_classes
+      class_names(
+        "govuk-pagination__link-title",
+        # "govuk-pagination__link-title--decorated" => block_mode?
```
peteryates, line 104: Not really sure what decoration is here or when it should be applied.

https://github.com/alphagov/govuk-frontend/pull/2610/files#diff-8f24600b8b58d8a6ff19d90cadc27de09b8314b4405af61913e4e509858d39a4R107

peteryates:
This PR is _mostly_ done now but there's a few tweaks that need to [be made upstream](https://github.com/alphagov/govuk-frontend/pull/2610) around alignment and whitespace. For example, from the guide preview (note how 'Chapter' and 'Edmund' aren't aligned vertically:

Screenshot from 2022-05-19 12-18-56

peteryates:
Alignment looks to be correct now thanks to the latest CSS updates on the `govuk-frontend` branch

Screenshot from 2022-06-11 20-06-47

peteryates:
It's pretty much ready, need to merge #337 and rebase. Should be out today or tomorrow!

PR #323, opened 2022-05-28
Switch to `html-attributes-utils`

peteryates:
This change shouldn't affect anything externally but now will use `html-attributes-utils` to do all the merging of HTML attributes and classes. It's also the first step towards dropping the separate `classes` array and relying on a single hash that can contain a custom classes key if required.


cpjmcquillan:
It's funny you raised this, because I was just looking at [this gem (and associated Rails PR)](https://github.com/seanpdoyle/action_view-attributes_and_token_lists) the other day and wondering whether it could be used here.

I haven't tried it out yet, but it might be something worth considering.

peteryates:
Ah thanks for linking to that @cpjmcquillan, hadn't seen it and the approach looks sound. Will definitely take a closer look.

[html-attributes-utils](https://github.com/DFE-Digital/html-attributes-utils) is just a standalone version of the functionality that was in the form builder, trying to take the pain out of converting space-separated lists of classes into arrays so they can be merged, plus getting rid of empty values.

tvararu (approved):

csutter:
In an unfortunate example of [Hyrum's law](https://www.hyrumslaw.com), we inherit from `GovukComponent::Base` in our components and rely extensively on overriding `default_classes`, so this has been a breaking change for us 😬 

This is on us for the most part, and too late to change versioning etc, but worth perhaps adding a warning to the release notes that consumers subclassing `GovukComponent::Base` may have to do some work?

peteryates:
@csutter damn sorry to hear that. I'll add a note. Let me know if you need a help updating stuff, I've done plenty over the last few weeks 😓

Issue #321, opened 2022-05-24
Section break helper


paulrobertlloyd:
:wave: Long time, no feature request!

There are currently no helpers for typography. Arguably this can/should remain the concern of a different gem (such as [GOV.UK Markdown renderer](https://github.com/dfE-Digital/govuk_markdown)).

However, it’s not unusual for [section breaks](https://design-system.service.gov.uk/styles/typography/#section-break) to be used in layouts, outside of content. My advice to colleagues regarding building services using this gem is “if you’re writing HTML, you probably shouldn’t be”. Using HTML for section breaks goes against that advice.

## Proposal

Rather than write this HTML

```html
<hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
```

provide a helper with options consistent with those used elsewhere:

```erb
<%= govuk_section_break %> =>
<hr class="govuk-section-break">

<%= govuk_section_break(visible: true) %> =>
<hr class="govuk-section-break govuk-section-break--visible">

<%= govuk_section_break(visible: true, size: "m") %> =>
<hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">

<%= govuk_section_break(visible: true, size: "s") %> =>
[This would throw an error as small section breaks are not a thing, weirdly]
```

cpjmcquillan:
This sounds sensible to me, and I'm happy to implement it.

@peteryates what do you think?

peteryates:
Yeah, good call @paulrobertlloyd - I suspect they're implemented by just about every service going.

Amazing, thanks @cpjmcquillan!

peteryates:
Might be of interest but I experimented a bit with some other [layout-related components](https://github.com/DFE-Digital/govuk-components/pull/52/files) a while ago. In the columns example there's a bit of extra complexity but there are a few things that could be made easier just with a helper, like rows, headings, lists.

Could be an interesting follow up to this work.

paulrobertlloyd:
Seeing as this issue is still open…

Tried upgrading to 3.0.5 but got an error:

`expected file /home/runner/work/submit-social-housing-lettings-and-sales-data/submit-social-housing-lettings-and-sales-data/vendor/bundle/ruby/3.1.0/gems/govuk-components-3.0.5/app/components/govuk_component/traits/custom_classes.rb to define constant GovukComponent::Traits::CustomClasses, but didn't`

[...]

PR #325, opened 2022-05-29
Add a section break component

cpjmcquillan:
In the Typography section of the design system, there is a part on the [Section
break element](https://design-system.service.gov.uk/styles/typography/#section-break).

This adds a component (and respective helper method) for rendering a section
break element, as suggested in #321.

The visibility and size of the component is configurable.

### DIFF
--- a/app/components/govuk_component/section_break_component.rb
+++ b/app/components/govuk_component/section_break_component.rb
@@ -1,0 +1,40 @@
+class GovukComponent::SectionBreakComponent < GovukComponent::Base
+  SIZES = %w(m l xl).freeze
+
+  def initialize(visible: false, size: nil, classes: [], html_attributes: {})
+    super(classes: classes, html_attributes: html_attributes)
+
+    @visible = visible
+    @size    = size
+  end
+
+  def call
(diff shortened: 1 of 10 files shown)


app/components/govuk_component/section_break_component.rb
```diff
@@ -0,0 +1,40 @@
...
+    ).split
+  end
+
+  def size_class
+    if size.blank?
+      ""
+    elsif size.in?(SIZES)
+      "govuk-section-break--#{size}"
+    else
+      raise ArgumentError, "invalid size #{size}, supported sizes are #{SIZES.to_sentence}"
```
cpjmcquillan, line 33: I'm not sure whether we should raise an error, or fallback and warn.

It looks like we raise errors elsewhere though, so I've maintained that behaviour for now.

peteryates (approved):
Spot on!

app/components/govuk_component/section_break_component.rb
peteryates: I think raising makes sense here, I'd want to find out immediately if there's an invalid size. Perhaps this is something that should be configurable though.

PR #326, opened 2022-05-30
Add missing attributes to #govuk_button_link_to

tvararu:
This was missing some of the attributes that the design system recommends for links that are styled as buttons:

https://design-system.service.gov.uk/components/button/start/index.html

User provided attributes still get `.merge`d on top, so this shouldn't break any existing overrides if someone has added these attributes manually. This is backwards compatible and is a bugfix or new feature depending on how you look at it.

### Screenshot from the deploy preview
image

### DIFF
--- a/app/components/govuk_component/start_button_component.rb
+++ b/app/components/govuk_component/start_button_component.rb
@@ -1,4 +1,10 @@
 class GovukComponent::StartButtonComponent < GovukComponent::Base
+  BUTTON_ATTRIBUTES = {
+    role: 'button',
+    draggable: 'false',
+    data: { module: 'govuk-button' }
+  }
+
   attr_reader :text, :href
 
   def initialize(text:, href:, classes: [], html_attributes: {})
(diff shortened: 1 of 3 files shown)


app/components/govuk_component/start_button_component.rb
```diff
@@ -1,4 +1,10 @@
 class GovukComponent::StartButtonComponent < GovukComponent::Base
+  BUTTON_ATTRIBUTES = {
+    role: 'button',
+    draggable: 'false',
+    data: { module: 'govuk-button' }
+  }
```
peteryates, line 6: Looks great overall, the only bit I'm not sure on is where `BUTTON_ATTRIBUTES` should live if it's shared between components and helpers.

peteryates, line 6:
```suggestion
  }.freeze
```

peteryates (approved):
Makes sense and looks good - thanks!

PR #335, opened 2022-06-18
Handle `classes: nil` more gracefully

### LINKED ISSUE #334: Deal with `nil` classes more gracefully
There was a recent report from Apply where a struct used for navigation items is initialised without a `classes` param it defaults to `nil`, and inside the gem we assume it's an array and [try to append to it](https://github.com/DFE-Digital/govuk-components/blob/main/app/components/govuk_component/header_component.rb#L80).

We really should ensure it's not nil before trying that!

peteryates:
When the classes param is `nil` the library fails while attempting to append conditional classes to it. The error message doesn't make it obvious what the problem is.

Here we are addressing it by emitting a warning when nil classes are detected and overwriting the nil with an empty array.

Fixes #334

### DIFF
--- a/app/components/govuk_component/base.rb
+++ b/app/components/govuk_component/base.rb
@@ -4,6 +4,11 @@ class GovukComponent::Base < ViewComponent::Base
   attr_reader :html_attributes
 
   def initialize(classes:, html_attributes:)
+    if classes.nil?
+      Rails.logger.warn("classes is nil, if no custom classes are needed omit the param")
+
+      classes = []
+    end
     # FIXME: remove first merge when we deprecate classes
     #
     # This step only needs to be here while we still accept classes:, now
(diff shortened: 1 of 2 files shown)


frankieroberto (approved):
Thanks for doing this.

Do you think we could explicitly allow `nil` to be set, rather than logging a warning? Could even just blindly call `to_a` on the variable, and then it’d work with anything that implements that, like Sets? Don’t know how Ruby-ish that would be though!

peteryates:
No problem. The change does do a `classes = []` after logging a warning when `classes.nil?`. The issue with doing a `.to_a` is that there'd need to be more logic to prevent it from being called on strings.

We could use [`Array.wrap`](https://www.rubydoc.info/docs/rails/3.1.1/Array.wrap) but I think erroring out when the param is anything except `String`, `Array` or `nil` is broadly in line with Rails' helpers.

PR #343, opened 2022-06-29
Allow start button to be a button

thomasleese:
There are certain situations where the start button should submit a form rather than taking the user to a new location. Effectively, this should render using `button_to` instead of `link_to`.

Idea came from this comment: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/174#discussion_r910155776

I can't get the tests to pass at the moment, something to do with them depending on the view context helpers, but I'm not familiar with view components.

### DIFF
--- a/guide/content/components/start-button.slim
+++ b/guide/content/components/start-button.slim
@@ -12,4 +12,22 @@ p
   caption: "Start button",
   code: start_button_normal)
 
+== render('/partials/example.*',
+  caption: "Rendering a start button in a form",
+  code: start_button_as_button) do
+
+  markdown:
+    If you need to submit data you can make the component render a form
+    containing a button instead of a link by passing `as_button: true`. This
+    approach uses the Rails `button_to` helper and will render a form that
+    will `POST` to the target URL.
+
+  div.govuk-warning-text
+    span.govuk-warning-text__icon aria-hidden="true"
+      | !
+    strong.govuk-warning-text__text
+      span.govuk-warning-text__assistive Warning
+      | The GOV.UK Design system
+      =< link_to("advises against submitting data with a start button", "https://design-system.service.gov.uk/components/button/#start-buttons").html_safe
+
 == render('/partials/related-info.*', links: start_button_info)
--- a/app/components/govuk_component/start_button_component.rb
+++ b/app/components/govuk_component/start_button_component.rb
@@ -1,29 +1,38 @@
 class GovukComponent::StartButtonComponent < GovukComponent::Base
   BUTTON_ATTRIBUTES = {
-    role: 'button',
     draggable: 'false',
     data: { module: 'govuk-button' }
   }.freeze
 
-  attr_reader :text, :href
+  LINK_ATTRIBUTES = BUTTON_ATTRIBUTES.merge({ role: 'button' }).freeze
 
-  def initialize(text:, href:, classes: [], html_attributes: {})
+  attr_reader :text, :href, :as_button
+
+  def initialize(text:, href:, as_button: false, classes: [], html_attributes: {})
     @text = text
     @href = href
+    @as_button = as_button
 
     super(classes: classes, html_attributes: html_attributes)
   end
 
   def call
-    link_to(href, **html_attributes) do
-      safe_join([text, icon])
+    if as_button
+      button_to(href, **html_attributes) do
+        safe_join([text, icon])
+      end
+    else
+      link_to(href, **html_attributes) do
+        safe_join([text, icon])
+      end
     end
   end
 
 private
 
   def default_attributes
-    BUTTON_ATTRIBUTES.merge({ class: %w(govuk-button govuk-button--start) })
+    (as_button ? BUTTON_ATTRIBUTES : LINK_ATTRIBUTES)
+      .merge({ class: %w(govuk-button govuk-button--start) })
   end
 
   def icon
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -68,7 +68,7 @@ module GovukLinkHelper
 
   def govuk_button_link_to(name = nil, options = nil, extra_options = {}, &block)
     extra_options = options if block_given?
-    html_options = GovukComponent::StartButtonComponent::BUTTON_ATTRIBUTES
+    html_options = GovukComponent::StartButtonComponent::LINK_ATTRIBUTES
       .merge build_html_options(extra_options, style: :button)
 
     if block_given?
--- a/guide/lib/examples/start_button_helpers.rb
+++ b/guide/lib/examples/start_button_helpers.rb
@@ -5,5 +5,11 @@ module Examples
         = govuk_start_button(text: "Start now", href: "#")
       START_BUTTON
     end
+
+    def start_button_as_button
+      <<~START_BUTTON
+        = govuk_start_button(text: "Start now", href: "#", as_button: true)
+      START_BUTTON
+    end
   end
 end
--- a/guide/lib/helpers/formatters.rb
+++ b/guide/lib/helpers/formatters.rb
@@ -11,6 +11,10 @@ module Helpers
     def variants
       []
     end
+
+    def protect_against_forgery?
+      false
+    end
   end
 
   module Formatters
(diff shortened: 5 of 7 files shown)


cpjmcquillan:
The [design system says this](https://design-system.service.gov.uk/components/button/) about the `Start button`.

> Start buttons do not submit form data, so they use a link tag rather than a button tag.

I don't agree with that statement though. If a button is semantically a button, then I think it should be a `<button>` and not an `<a>` tag —so I think this is a change worth making. Perhaps the documentation should be updated upstream too.

This change looks fine, but we might need to change some of the `html_attributes` when we render the button (e.g `role="button"`).

I haven't had a chance to look at the specs yet, but I will try and take a look soon.

Thanks for your contribution!

thomasleese:
> The [design system says this](https://design-system.service.gov.uk/components/button/) about the `Start button`.
> 
> > Start buttons do not submit form data, so they use a link tag rather than a button tag.

Ah you're right, I wasn't aware of that. Our needs behind making this a form is so that we can perform an action in our backend when a user starts the service, and it should be a `POST` since it's not idempotent.

> This change looks fine, but we might need to change some of the `html_attributes` when we render the button (e.g `role="button"`).

Good point, I've amended my commit to remove that unnecessary role.

> I haven't had a chance to look at the specs yet, but I will try and take a look soon.

I still can't seem to get them passing, I've tried custom controllers and view contexts. The issue seems to come from the fact that rendering a form expects there to be a CSRF token available, via `form_authenticity_token` which tries to access the `request.session` which seems to be `nil` (except it shouldn't be `nil`: https://github.com/github/view_component/blob/8dd9063d0b74125c0e20222ee76e64faac6d13a2/lib/view_component/test_helpers.rb#L93).
> 
> Thanks for your contribution!

peteryates:
Thanks for the PR @thomasleese. I had a quick dig and it's failing in the depths of actionview when trying to generate an `form_authenticity_token`

https://github.com/rails/rails/blob/147557dc16a5bd8bab96f999f359a570232e1873/actionview/lib/action_view/helpers/url_helper.rb#L729-L730

I think mocking this in RSpec _should_ allow tests to be written. This does indicate that the start button should _perhaps_ be a helper rather than a fully-fledged component though.

thomasleese:
> Thanks for the PR @thomasleese. I had a quick dig and it's failing in the depths of actionview when trying to generate an `form_authenticity_token`
> 
> I think mocking this in RSpec _should_ allow tests to be written. This does indicate that the start button should _perhaps_ be a helper rather than a fully-fledged component though.

Yeah, I wasn't sure about mocking as I didn't want to mock too much and find it doesn't actually work. What's the difference between a helper and a component?

peteryates:
@thomasleese there's not really much of a difference for the most part, just the helpers are vanilla Rails helper modules and the components require ViewComponent. However, what's available to the specs with type `:component` and `helper` does differ - we have lots of helper specs that do `button_to` without needing any extra setup.

I'm slightly confused because in the dummy app [forgery protection is disabled](https://github.com/DFE-Digital/govuk-components/blob/main/spec/dummy/config/environments/test.rb#L31). I thought that would prevent this kind of thing from happening 🤔 

Either way, we can simply override it in the spec which makes most of the tests pass:

```patch
diff --git a/spec/components/govuk_component/start_button_component_spec.rb b/spec/components/govuk_component/start_button_component_spec.rb
index 55a5c3b..2fe45a6 100644
--- a/spec/components/govuk_component/start_button_component_spec.rb
+++ b/spec/components/govuk_component/start_button_component_spec.rb
@@ -1,6 +1,11 @@
 require 'spec_helper'
 
 RSpec.describe(GovukComponent::StartButtonComponent, type: :component) do
+  before do
+    allow_any_instance_of(GovukComponent::StartButtonComponent)
+      .to(receive(:protect_against_forgery?).and_return(false))
+  end
+
   let(:component_css_class) { 'govuk-button--start' }
   let(:text) { 'Department for Education' }
   let(:href) { 'https://www.gov.uk/government/organisations/department-for-education' }

```

thomasleese:
> I'm slightly confused because in the dummy app [forgery protection is disabled](https://github.com/DFE-Digital/govuk-components/blob/main/spec/dummy/config/environments/test.rb#L31). I thought that would prevent this kind of thing from happening 🤔

@peteryates Yeah, I noticed that too and didn't really understand why it wasn't preventing it.

> Either way, we can simply override it in the spec which makes most of the tests pass:

I've amended the commit with the mocking, and the tests are all passing now! There's an issue with Semgrep, but I think that might be something to do with this PR coming from a fork, as it seems to be trying to fetch the branch from origin.

guide/content/components/start-button.slim
```diff
@@ -12,4 +12,8 @@ p
   caption: "Start button",
   code: start_button_normal)
 
+== render('/partials/example.*',
+  caption: "Start button as a submit button",
+  code: start_button_as_button)
```
peteryates, line 17: I think we could benefit from a sentence here on why one might want to use a form/button here rather than just a link.

Not sold on this but as a starting off point

```suggestion
  code: start_button_as_button) do

  markdown:
    If you need to submit data you can make the component render a form
    containing a button instead of a link by passing `as_button: true`. This
    approach uses Rails' `button_to` helper and will render a form that will
    `POST` to the target URL.

  div.govuk-warning-text
    span.govuk-warning-text__icon aria-hidden="true"
      | !
    strong.govuk-warning-text__text
      span.govuk-warning-text__assistive Warning
      | The GOV.UK Design system
      =< link_to("advises against submitting data with a start button", "https://design-system.service.gov.uk/components/button/#start-buttons").html_safe
```

guide/content/components/start-button.slim
```diff
@@ -12,4 +12,8 @@ p
   caption: "Start button",
   code: start_button_normal)
 
+== render('/partials/example.*',
+  caption: "Start button as a submit button",
```
peteryates, line 16: Without repeating the word button maybe?

```suggestion
  caption: "Rendering a start button in a form",
```

guide/content/components/start-button.slim
peteryates: It's also probably worth adjusting the opening sentence above so it says something more like:

> Start buttons don't usually submit form data, so they use a link tag rather than a button tag.

thomasleese: I've opened https://github.com/alphagov/govuk-design-system/pull/2238 to change the wording.

thomasleese:
@peteryates Thanks, I've amended the examples with your suggestions! I think I'll take the PR out of draft now.

peteryates:
Great, thanks for this @thomasleese. I'm happy with the changes, looking good to me.

paulrobertlloyd:
> The [design system says this](https://design-system.service.gov.uk/components/button/) about the `Start button`.
>
> > Start buttons do not submit form data, so they use a link tag rather than a button tag.

I was going to mention this, but sounds like there are valid use cases.

This is probably another area within the Design System which doesn’t account for how most services tend to work these days. That said, there’s an argument to be made that start pages – and therefore start buttons – should only appear on mainstream GOV.UK pages. That makes their inclusion in the Design System a bit questionable.

Maybe some guidance in this gems docs could reinforce this fact? Worry is that start buttons get used to submit ‘classic’ forms, which is a no-no.

paulrobertlloyd:
> Thoughts on the param name as_button

Works for me. Will just point out that the Design System uses an `element` param on its button component. That said, the start button component in this library is a bit of a special case already, so as long as consistent with other options/Rails equivalents, probably fine.

peteryates:
Yeah - I considered suggesting having the builder just output a `<button>` element and leaving the surrounding form as an exercise for the dev above - was never the biggest fan of `button_to` in Rails generally.

But, this way the button is 'protected' from being used to submit any data so I'm happy for it to go in.

thomasleese:
> Yeah - I considered suggesting having the builder just output a `<button>` element and leaving the surrounding form as an exercise for the dev above - was never the biggest fan of `button_to` in Rails generally.

Yeah, my plan initially was to make this code produce just a single button and allow the user to build the form around it, but I agree that it makes sense in this context with a start button to generate the form as well, and it means the start button component effectively behaves the same whether `as_button` is `true` or `false` - it takes the user to a href, one via `GET` and one via `POST`.

peteryates (approved):
Looks good to me.

PR #354, opened 2022-08-05
Require visually_hidden_text to be set on summary list actions

frankieroberto:
Visually-hidden text is nearly always required on the action links for summary list items, as otherwise they will often all just contain "Change", which is unhelpful for screen reader users who may hear links read out out-of-context of the rest of the page.

Unfortunately, it’s relatively common to forget to set this, and for this not to be picked up when reviewing code or the service. On the Apply for teacher training service, despite us being aware of this issue, several instances slipped through our review processes and were not picked up until an external accessibility audit was done.

This change aims to make this accessibility issue less common by requiring `visually_hidden_text` to be set on the action items of summary list rows. If it is not set, ruby will throw an error.

In the rare occasions where visually-hidden text is not required (say for instance there's only a single action link on the page, or all of the links have different link text), `visually_hidden_text` can be explicitly set to `nil`.

This would probably have do be released as a breaking change in the next major version release?

### DIFF
--- a/app/components/govuk_component/summary_list_component/action_component.rb
+++ b/app/components/govuk_component/summary_list_component/action_component.rb
@@ -1,7 +1,13 @@
 class GovukComponent::SummaryListComponent::ActionComponent < GovukComponent::Base
+  class VisuallyHiddenDefaultNilClass < NilClass; end
+
   attr_reader :href, :text, :visually_hidden_text, :attributes, :classes
 
-  def initialize(href: nil, text: 'Change', visually_hidden_text:, classes: [], html_attributes: {})
+  def initialize(href: nil, text: 'Change', visually_hidden_text: VisuallyHiddenDefaultNilClass, classes: [], html_attributes: {})
+    if config.require_summary_list_action_visually_hidden_text && visually_hidden_text == VisuallyHiddenDefaultNilClass
+      fail(ArgumentError, "missing keyword: visually_hidden_text")
+    end
+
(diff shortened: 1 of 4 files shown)


edwardhorsford:
Nice idea Frankie!

I've also had times where I didn't need visually hidden text because the visible text was already descriptive and unique.

Do we have control of the error message?

frankieroberto:
> I've also had times where I didn't need visually hidden text because the visible text was already descriptive and unique.

Interesting! I'd assumed this was really rare (for summary list actions). Do you have any examples to hand?

> Do we have control of the error message?

Ruby will throw a relatively-descriptive "missing argument error". I don't think there’s any easy way to customise this though. Hopefully it'd be enough?

Screenshot 2022-08-05 at 15 27 58

peteryates:
Looks good, thanks @frankieroberto.

I agree with making it opt out rather than opt in, the cases where you don't need hidden text are certainly rare.

Agree it's going to be a breaking change. Unlike the form builder this lib doesn't support configuration so merging it in a 'disabled by default' state isn't an option yet.

frankieroberto:
@peteryates great - we could leave this PR open until we're ready for the next breaking change release? (Do we have any other breaking changes on the horizon?)

peteryates (approved):
Now #355 is merged and we've appended @josephhull676's config option here I'm happy for this to go in.

We'll leave it disabled by default for now but toggle it at the next major release.

PR #355, opened 2022-08-05
Make the library configurable

peteryates:
This is very much a WIP look at adding confguration to the gem. It's a useful feature to have because it will allow for some preferences to be set across entire apps, rather than repeatedly when using components. It also makes it easier for us to bring in breaking changes by allowing them to be introduced in a disabled state and toggle the default at the next major release (see #354).

* [x] identify the things that should be configurable, things like:
  * default button text
  * default titles
  * headings
  * sizes
  * widths
* [x] add some tests covering each config option
* [x] cover the functionality in the guide

### Back link
* [x] `text`

### Breadcrumbs
* [x] `hide_in_print`
* [x] `collapse_on_mobile`

### Cookie banner
* [x] `aria_label`
* [x] `hide_in_print`

### Footer #358
* [x] `copyright`
* [x] `meta_licence`
* [x] `meta_text`

### Header
* [x] `logotype`
* [x] `menu_button_label`
* [x] `navigation_label`
* [x] `service_url`
* [x] `service_name`

### Notification banner
* [x] `title_id`
* [x] `disable_auto_focus`
* [x] `title_heading_level`
* [x] `success`

### Pagination

* [x] `landmark_label`
* [x] `next_text`
* [x] `previous_text`

### Phase banner

* [x] `tag`
* [x] `text`

### Section break

* [x] `visible`
* [x] `size`

### Start button

* [x] `as_button`

### Summary list

* [x] `borders`
* [x] ~~`actions`~~ (omitted this as I'm not sure it makes sense to be configurable)

### Tag

* [x] `colour`

### Warning text

* [x] `icon_fallback_text`
* [x] `icon` (_not currently configurable_)


cpjmcquillan:
I think this is a really nice idea. 🪄 

Primer ViewComponents defines a [`fetch_or_fallback` helper method](https://cs.github.com/primer/view_components/blob/4227e457cac9602d7848aaa00ebde7424bb1a7bb/app/lib/primer/fetch_or_fallback_helper.rb?q=fetch_or_fallback#L28) that might be interesting reading. 

You might have seen it already, and It's probably a bit too involved for a first pass, but I thought it might be useful nonetheless!

peteryates:
@cpjmcquillan thanks for that Connor. `#fetch_or_fallback` is such a good idea - way more graceful than doing it ad hoc. Here and in the form builder we're relying on setting default arguments for keyword args, but I think in the helpers (where they're mimicking Rails' method signatures much more closely) it would be a great fit. I'll have a look into using a similar pattern.

PR #356, opened 2022-08-10
Upgrade govuk-frontend to 4.3.0

peteryates:


frankieroberto:
It looks like as part of this we should add the `hidden` attribute to the menu button on [line 40 of the header component](https://github.com/DFE-Digital/govuk-components/blob/main/app/components/govuk_component/header_component.html.erb#L40)?

See [upstream change in govuk-frontend](https://github.com/alphagov/govuk-frontend/pull/2727/files#diff-dcc9d461905ec3f2d5c9a73c4c7bc09499b0c8265fe7e32c097b0e6649e1106fL64).

frankieroberto (approved):

PR #358, opened 2022-08-12
Make footer component configurable

JonF1:
Building on https://github.com/DFE-Digital/govuk-components/pull/355/, this pr allows the meta_text, copyright_text and copyright_url attributes of the footer component to be configurable


peteryates (approved):
Looks great!

PR #359, opened 2022-08-12
Make breadcrumbs component configurable

robyoung91:
This PR allows `hide_in_print` and `collapse_on_mobile` in the breadcrumbs component to be configurable.


peteryates (approved):
Looks great.

PR #362, opened 2022-08-19
Make phase banner configurable

JonF1:
Allows the tag and text attributes of the phase banner component to be configurable

### DIFF
--- a/spec/components/govuk_component/configuration/phase_banner_component_configuration_spec.rb
+++ b/spec/components/govuk_component/configuration/phase_banner_component_configuration_spec.rb
@@ -1,0 +1,30 @@
+require 'spec_helper'
+
+RSpec.describe(GovukComponent::PhaseBannerComponent, type: :component) do
+  describe "configuration" do
+    after { Govuk::Components.reset! }
+
+    describe 'default phase banner component tag and text' do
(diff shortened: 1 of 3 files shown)


spec/components/govuk_component/configuration/phase_banner_component_configuration_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      let(:overridden_default_text) { 'Public service' }
+
+      before do
+        Govuk::Components.configure do |config|
+          config.default_phase_banner_component_tag = overridden_default_tag
+          config.default_phase_banner_component_text = overridden_default_text
+        end
+      end
+
+      subject! { render_inline(GovukComponent::PhaseBannerComponent.new(**kwargs)) }
```
peteryates, line 22: I think it makes sense to drop `kwargs` here, and above (lines 4-6), they're not actually used for anything.

peteryates (approved):
Looks great to me. Left one comment on removing a few variables we perhaps don't need.

peteryates:
Nice, good to go! Thanks @JonF1

JonF1:
> Nice, good to go! Thanks @JonF1

No problem 👍

PR #363, opened 2022-08-21
Support links opening in new tab

### LINKED ISSUE #360: Idea: make it easier to create links that open in new windows
[According to the design system,](https://design-system.service.gov.uk/styles/typography/#opening-links-in-a-new-tab) links that open in new tabs/windows need to:

* have `target: "_blank"`
* be `rel: "noreferrer noopener"` (to prevent [reverse tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing))
* include a description that informs the user the window will open in a new window, usually `(opens in new tab)` or similar

Rather than having to remember all of these things every time, we could bake the functionality into the library. It could be implemented as:

* a new helper - perhaps `#govuk_external_link_to`
* via a keyword argument on `#govuk_link_to` and friends - maybe `external: true`

The text appended to the end should be `nil` by default (leaving it to the dev) but can be defaulted so it always adds the `(link opens in new tab)` suffix - unless supressed with `external_text: nil` or similar.

peteryates:
This PR adds new keyword argument `new_tab` to `#govuk_link_to` and `#govuk_button_link_to` that makes it easier (and safer) for developers to create links that open in a new tab/window.

The design system specifies that the attributes `rel="noreferrer noopener" target="_blank"` should be present to prevent [reverse tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing). This is often forgotten by developers.

The default value for `new_tab` is false. The new behaviours are as follows:

* when `new_tab: false` - no change, the helper works in the normal fashion
* when `new_tab: true` - the default suffix is added after the link text
* when `new_tab: "a string"` - the supplied suffix is added after the link text

The default suffix is "(opens in new tab)".

Fixes #360

### DIFF
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -1,4 +1,8 @@
+require "html_attributes_utils"
+
 module GovukLinkHelper
+  using HTMLAttributesUtils
+
   LINK_STYLES = {
     inverse:          "govuk-link--inverse",
     muted:            "govuk-link--muted",
@@ -95,10 +99,11 @@ private
   def build_html_options(provided_options, style: :link, new_tab: false)
     element_styles = { link: LINK_STYLES, button: BUTTON_STYLES }.fetch(style, {})
 
-    remaining_options = {
-      **remove_styles_from_provided_options(element_styles, provided_options),
-      **new_tab_options(new_tab)
-    }
+    # we need to take a couple of extra steps here because we don't want the style
+    # params (inverse, muted, etc) to end up as extra attributes on the link.
+
+    remaining_options = new_tab_options(new_tab)
+      .deep_merge_html_attributes(remove_styles_from_provided_options(element_styles, provided_options))
 
     style_classes = build_style_classes(style, extract_styles_from_provided_options(element_styles, provided_options))
 
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -13,6 +13,11 @@ module GovukLinkHelper
     warning:   "govuk-button--warning",
   }.freeze
 
+  NEW_TAB_ATTRIBUTES = {
+    target: "_blank",
+    rel: "noreferrer noopener"
+  }.freeze
+
   def govuk_link_classes(*styles, default_class: 'govuk-link')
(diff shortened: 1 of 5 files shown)


frankieroberto:
Looks good! Might be worth checking if this plays nicely if you were to also specify those `rel` and `target` attributes as options (even though this seems unlikely)?

eg:

```ruby
govuk_link_to "Blah", "https://www.example.com", {rel: "help", target: "new"}, new_tab: "true"
```

Behaviour here is a bit unpredictable, but possibly the ideal would be to:

* append `noreferrer noopener` to the `rel` value given, separated by a space
* the `target` value given manually should override the `_blank` one (or possibly the other way round, hard to know...)

peteryates:
> Behaviour here is a bit unpredictable, but possibly the ideal would be to:
>
> * append noreferrer noopener to the rel value given, separated by a space
> * the target value given manually should override the _blank one (or possibly the other way round, hard to know...)

This should be quite straightfowrard since [this change to html-attributes-utils](https://github.com/DFE-Digital/html-attributes-utils/commit/41c1eef97d96a4b2e44c6aeaf304ca515020a531).

Unfortunately as we can see from the builds, Ruby 3.0.0 [changed the way keyword arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) meaning that the approach here needs to be adjusted.

I'm not sure it's possible to change without breaking Rails compatibility or adding extra nested args.

peteryates:
Going to hold off on this until Ruby `3.2.0` is released at Christmas and support for Ruby `2.7` is dropped.

peteryates:
This PR can  follow on from #381 and will be in the upcoming release  🥳

peteryates:
> Behaviour here is a bit unpredictable, but possibly the ideal would be to:
>
>  *  append noreferrer noopener to the rel value given, separated by a space
>  *  the target value given manually should override the _blank one (or possibly the other way round, hard to know...)

This behaviour is now included and covered by [these tests](https://github.com/DFE-Digital/govuk-components/blob/9570c4cab0c7e450dc87d1d018d7fd2e6ca56642/spec/helpers/govuk_link_helper_spec.rb#L139).

PR #367, opened 2022-09-07
When `visually_hidden_text` is omitted from summary list actions, don't render anything

peteryates:
This fixes a bug where when `visually_hidden_text` is totally omitted the `VisuallyHiddenDefaultNilClass` class name is rendered in the visually hidden span. It turns out you can't simply inherit from `NilClass`.

We can use `false` here instead.

### DIFF
--- a/app/components/govuk_component/summary_list_component/action_component.rb
+++ b/app/components/govuk_component/summary_list_component/action_component.rb
@@ -1,18 +1,17 @@
 class GovukComponent::SummaryListComponent::ActionComponent < GovukComponent::Base
-  class VisuallyHiddenDefaultNilClass < NilClass; end
-
   attr_reader :href, :text, :visually_hidden_text, :attributes, :classes
 
-  def initialize(href: nil, text: 'Change', visually_hidden_text: VisuallyHiddenDefaultNilClass, classes: [], html_attributes: {})
-    if config.require_summary_list_action_visually_hidden_text && visually_hidden_text == VisuallyHiddenDefaultNilClass
+  def initialize(href: nil, text: 'Change', visually_hidden_text: false, classes: [], html_attributes: {})
+    @visually_hidden_text = visually_hidden_text
+
+    if config.require_summary_list_action_visually_hidden_text && visually_hidden_text == false
       fail(ArgumentError, "missing keyword: visually_hidden_text")
     end
 
     super(classes: classes, html_attributes: html_attributes)
 
-    @href                 = href
-    @text                 = text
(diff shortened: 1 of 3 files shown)


spec/components/govuk_component/configuration/summary_list_configuration_spec.rb
```diff
@@ -25,18 +25,37 @@
...
+          render_inline(GovukComponent::SummaryListComponent.new) do |sl|
+            sl.row do |row|
+              row.key(text: "key one")
+              row.value(text: "value one")
+              row.action(text: "action one", href: "/action-one", visually_hidden_text: visually_hidden_text)
+            end
           end
         end
+
+        specify "raises an error when no visually hidden text is supplied" do
```
cpjmcquillan, line 40:
```suggestion
        specify "renders a span with the visually hidden text" do
```

peteryates: Thanks @cpjmcquillan - rushing and failing 😩

app/components/govuk_component/summary_list_component/action_component.rb
```diff
@@ -1,18 +1,17 @@
...
-  class VisuallyHiddenDefaultNilClass < NilClass; end
-
   attr_reader :href, :text, :visually_hidden_text, :attributes, :classes
 
-  def initialize(href: nil, text: 'Change', visually_hidden_text: VisuallyHiddenDefaultNilClass, classes: [], html_attributes: {})
-    if config.require_summary_list_action_visually_hidden_text && visually_hidden_text == VisuallyHiddenDefaultNilClass
+  def initialize(href: nil, text: 'Change', visually_hidden_text: false, classes: [], html_attributes: {})
+    @visually_hidden_text = visually_hidden_text
+
+    if config.require_summary_list_action_visually_hidden_text && visually_hidden_text == false
```
cpjmcquillan, line 7: What happens if this config is enabled, but someone renders the component with `nil` or `""` as the visually hidden text?

peteryates: `nil` or `""` can be used to override the requirement as there might be some places where visually hidden text isn't necessary

cpjmcquillan: Ok cool, that makes sense. 

I wasn't sure if the intention of the config setting was _we should never render this component without visually hidden text_, or if it was _please provide visually hidden text unless you're omitting it intentionally_.

cpjmcquillan (approved):

PR #368, opened 2022-09-20
Stop using GDS Transport and the crown logo

peteryates:
It was brought to my attention that the components guide [doesn't follow the rules for community resources and tools](https://design-system.service.gov.uk/community/contribution-criteria/#developing-a-community-resource-or-tool) with regards to fonts and the use of the crown logo.

This puts the library in a slightly difficult situation because it's raison d'être is to make building apps that follow the Design System easier.

## Changes

image

### Logos

* The crown logo has been removed and the service name has been changed to 'Ruby on Rails components'.
* GOV.UK has been left on. [A previous iteration](https://user-images.githubusercontent.com/128088/191212889-a54dd4e2-dee9-4ca7-9fdf-b9d69f16fd06.png) tried replacing GOV.UK with Department for Education, but it caused confusion ("are these DfE-only components?")
* The Crown Copyright logo has been removed from the main footer. **It has not been removed from footer component examples**.

### Colours

The blue that's used on lots of GOV.UK services has been switched for the X-GOV blue. This should provide a visual indication that it's not an official project. 

### Font changes

On [the typography front](https://design-system.service.gov.uk/styles/typography/), the relevant section of the Design System docs states:

> If your service is publicly available on a subdomain other than service.gov.uk, use an alternative typeface like Helvetica or Arial.

Neither of the suggested fonts look like GDS Transport. Not looking like the thing we're building for would be a huge failure of the guide; it might even make people think they're incompatible.

Instead my company (Graphia Ltd), has bought licences for [New Transport](http://www.newtransport.co.uk/) webfonts, the font GDS Transport is based on. While not exact, they are similar enough that the documentation makes sense.

### Thoughts

This seems to be an inconsistently-enforced rule. There are plenty of other projects that use the font and logos on non `service.gov.uk` domains. Some examples include:

* [The GOV.UK Prototype kit](https://govuk-prototype-kit.herokuapp.com/docs)
* [GOV.UK Design System review apps](https://deploy-preview-2251--govuk-design-system-preview.netlify.app/)


frankieroberto:
Agree that this rule isn’t well enforced - there’s loads of examples out there.

This feels like a reasonable compromise to me though, and I like the name change in the black header - makes it a bit clearer that this is for Ruby on Rails. Might also be worth adding Rails to the h1 on the homepage?

> Build your GOV.UK service using Rails components

tvararu (approved):

cpjmcquillan (approved):
Not even being allowed to use the font seems a little extreme!

But I think your proposal is a reasonable solution, thank you ⭐

PR #372, opened 2022-11-01
Eject from rubocop govuk

adipasquale:


adipasquale:
sorry wrong repo again 🤦

PR #373, opened 2022-11-03
Make `view_component` version less specific

peteryates:
We've seen a problem arise a few times when new versions of `view_component` are released and the automatic upgrades resolve `govuk-components` back to version `1.2.0`.

Now `view_component` has stabilised we can loosen it.


peteryates:
Guide is failing to build, looks related to [this issue](https://github.com/ViewComponent/view_component/issues/1565).

PR #387, opened 2023-01-01
Add table footer support to table component

peteryates:
This is an experiment and builds on #382.

e407db5b6a2ba adds a [tfoot](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tfoot) component that can be rendered inside tables. It works like a body component but there can be only one. It's rendered after all the other slots.

## Consdierations

Supporting footers in the automatic modes (when passing in arrays of arrays) is probably going to be a bit of a pain, the component would have to do the tallying. It _could_ be supported by adding a `foot` param that (like the head param) sets the values directly, leaving any calculations up to the developer.


frankieroberto:
Looks good! I think in theory (if I understand the specs right), the "Total" cell in the footer should be a `th` with `scope=row` and the actual total cells should be `td` elements with no scope (even though they summarise the cells above)?

peteryates:
@frankieroberto yeah, I have a couple of `x`'d specs to cover those scenarios. I need to do a little reworking to separate out the idea of being a `th` and a `govuk-table__header` - they were previously treated being the same. It's starting to come together but needs some refinement.

Issue #390, opened 2023-01-24
Version 4.0.0


peteryates:
# Version 4.0.0

* Upgrade to ViewComponent 3.0.0rc6
* Upgrade to [govuk-frontend 4.5.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.5.0)
* Improve the table component

  * Add support for table header scopes #382

    [Table header scopes](https://www.w3.org/WAI/EO/Drafts/tutorials/tables/scope/) are used to identify cells that are covered by a header cell.
    They can be manually added using the `scope` parameter but will default to `col` for
    table headings in the `<thead>` and to `row` for table headings in the `<tbody>`.

  * Add support for table footers #387

    [Table footers](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tfoot) allow us to summarise a table's contents. Often they are used to
    show numeric totals.

  * Add support for `<colgroup>` and `<col>` elements #393

    [Column groups](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/colgroup) define groups of columns within a table. They are useful when single
    categories of data need to span multiple columns

  There are examples on the new functionality on [the table page of the guide](https://govuk-components.netlify.app/components/table)

* Move the space between the visible and visually hidden text in summary list actions inside the visually hidden span. #408 
   
   The markup will change in the following manner:

   ```diff
         <a class="govuk-link" href="#">
   -       Change <span class="govuk-visually-hidden">age</span>
   +       Change<span class="govuk-visually-hidden"> age</span>
         </a>
   ```

* Add [Summary Cards](https://design-system.service.gov.uk/components/summary-list/#summary-cards) #351

  Summary cards provide visual separation when displaying more than one summary list
  on a page. They can also contain links to actions that affect the card as a whole.

  There are examples on the new functionality on [the summary list page of the guide](https://govuk-components.netlify.app/components/summary-list/#summary-cards)

* Make Summary Card `action` argument override the default #413
  
   Previously the action column would be shown if there are actions present regardless of what's set in the `actions` argument. This meant that it was impossible to suppress the actions column. Now it works in a more sensible fashion and the overridden value is used when set. Thank you @vassyz for contributing this one.

* Upgrade to the latest versions of Ruby and Rails.

  * The supported Ruby versions are: 3.0.5, 3.1.3 and 3.2.1
  * The supported Rails versions are: 7.0.4 and 6.1.7

### New view component slot syntax

[...]

PR #398, opened 2023-02-05
Drop the `classes` keyword

peteryates:
Since releasing [html-attributes-utils](https://github.com/DFE-Digital/html-attributes-utils) last year we've had two ways to set classes on components, via:

1. `classes: 'xyz'`
2. `html_attributes: { class: 'xyz' }`

html_attributes is preferred because:

* it means the `class` parameter name matches the final HTML attribute
* all attributes are dealt with in the same way, simplifying the code
* a single approach is easier to document and understand

### Final steps

* [ ] get some feedback on this change
* [ ] fix the remaining `classes` style parameters
```bash
      ag "\w+_classes" app/components/ --only-matching --nofilename --nobreak | sort | uniq -c | sort -nr
        6 custom_container_classes
        5 meta_classes
        5 container_classes
        4 default_classes
        3 custom_navigation_classes
        2 title_classes
        2 navigation_classes
        2 link_classes
        2 li_classes
        2 label_classes
        2 custom_meta_classes
        1 govuk_link_classes
```
* [ ] write release notes for #390


frankieroberto:
The downside to this is that `html_attributes: { class: 'xyz' }` is longer, and `classes` are probably the mostly commonly set attribute?

Would it be possible to allow html attributes to be set directly as top-level keyword arguments (basically any arguments not recognised by the component are assumed to be html attributes)? That might be more similar to how Rails’s view helpers work?

peteryates:
Yeah - it's a tricky one. Accepting addition arguments work well providing there's no clash with the method params.

I think having a single approach for all attributes is easier to understand, support and document - but I can see how it's a touch less convenient in some circumstances.

I'll post a thread on Slack to see if people have strong opinions.

frankieroberto:
Might be worth checking how many times `classes:` is used in the various services using this Gem?

For Apply for teacher training, there seems to be [66 usages](https://github.com/DFE-Digital/apply-for-teacher-training/search?q=classes).

paulrobertlloyd:
For the Design System Nunjucks components, you can often set values that could conceivably be set via the `attributes` option. For example, for the text input component, you can set `type`, `inputmode`, `spellcheck` and `autocomplete` options. You can also set a `classes` option.

The model therefore appears to be that common attributes are bought to the top, both to reduce verbosity, but also I suspect increase the visibility of these options.

I think this is a good model to follow. As @frankieroberto pointed out, there are quite often cases where additional classes need to be added to a component in order to modify it; requiring a nested option introduces complexity and verbosity.

TLDR; I'd be tempted to leave things as they are. `classes` is fine, especially as it's the same option name used in the design system. As to the code being more complex, I’d refer to the GOV.UK Design Principles: do the hard things to make things easy for users 😉

paulrobertlloyd:
So in fact, I’d be keen on going in an opposite direction... perhaps by ensuring the attributes that are bought to the top level in the Design System components should also be bought to the top in this project (if they aren’t already).

frankieroberto:
side note on `class` vs `classes`: I presume it's singular in HTML as it was originally intended to only have a single class, and the idea of space-separated classes came later but it was too late to change the attribute name? And then maybe the design system team went with the plural version in the nunjucks macro to make it clearer that you can add multiple? Either way, the inconsistency is confusing but I guess something we have to live with? 🤷

peteryates:
[Back in the day](https://www.w3.org/MarkUp/html3/html3.txt) classes were hierarchical 😅 

> For instance, `<H2 CLASS=Section>` defines a level 2 header that acts as a
> section header. By convention, **the class names are interpreted
> hierarchically, with the most general class on the left and the most specific
> on the right**, where classes are separated by a period.

peteryates:
Closing this for now. While it's clear `html_attributes` makes the gem cleaner internally lots of people are using `classes`. Perhaps this will be revisited at some point in the future.

Issue #402, opened 2023-02-27
SummaryList output for screenreaders


paulodeon:
We had some feedback from the Digital Accessibility Centre regarding the output of `GovukComponent::SummaryListComponent::ActionComponent`

> Some link text was formatted in a way that weas difficult for screen reader users to  understand. 
>
> The ‘change’ link within the ‘Your name’ row within the summary list is formatted such that  screen reader software does not reliably interpret it. The space between the link text and  the hidden text is not recognised by some screen reader software, and thus will not be  announced properly to screen reader users. For example, ‘Change your name’ is interpreted  as ‘Changeyour name’.
> 
> Current output
```html
<a class="govuk-link" href="/referrals/3/referrer 
name/edit?return_to=%2Freferrals%2F3%2Freferrer"> 
Change <span class="govuk-visually-hidden">your name</span> 
</a> 
```
> 
> Solution
> 
> For reliable screen reader parsing of link text, we recommend placing a non-breaking space  character in the visually-hidden text using the HTML entity “&nbsp;”. Placing the non breaking space within the visually hidden text prevents a space from being placed at the end  of the visible link. The trailing space before the visually hidden <span> can then be removed  to prevent some browsers from interpreting two spaces adjacent to each other. 
>
> 
```html
<a class="govuk-link" href="/referrals/3/referrer 
name/edit?return_to=%2Freferrals%2F3%2Freferrer"> 
Change<span class="govuk-visually-hidden">&nbsp;your name</span> </a>
```

Is this something we'd accept into the library? 

Happy to create a PR if so

peteryates:
Ah, interesting!

Yeah I'd be more than happy to accept it - having to do it manually definitely feels a bit clunky. I think it should at least be a [global config option](https://github.com/DFE-Digital/govuk-components/blob/main/lib/govuk/components/engine.rb#L102), but perhaps should be an argument too?

frankieroberto:
This is also being tracked upstream here by the Design System team: https://github.com/alphagov/govuk-frontend/issues/2547

peteryates:
I think this is solved now.

PR #403, opened 2023-03-01
Add configuration option for visually hidden space on SummaryListComponent::ActionComponent

paulodeon:
As per issue #402. We are adding a configuration option to allow moving the space in the summary list actions from the visually shown to visually hidden text.

I'm not sure whether we should be using nbsp or a regular space here but I've put in nbsp for now as that was the recommendation

### DIFF
--- a/app/components/govuk_component/summary_list_component/action_component.rb
+++ b/app/components/govuk_component/summary_list_component/action_component.rb
@@ -19,8 +19,10 @@ class GovukComponent::SummaryListComponent::ActionComponent < GovukComponent::Ba
   end
 
   def call
+    space = config.summary_list_action_visually_hidden_space ? nil : " "
+
     link_to(href, **html_attributes) do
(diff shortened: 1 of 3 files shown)


frankieroberto:
I'm not sure this needs to be a configurable option... If it's needed we should always use it.

That said, the [upstream issue](https://github.com/alphagov/govuk-frontend/issues/2547) suggests that it might be an issue even with the non-breaking space.

In the short term, could we do the small step of moving the existing space inside the `<span class="govuk-visually-hidden">` rather than outside it? (This would bring the Rails component in alignment with the nunjucks version.)

peteryates:
PR looks good, as @frankieroberto let's hold fire a while and see what happens upstream. Many thanks for submitting @paulodeon

peteryates:
Closing this now as it seems shifting the space has done the trick.

PR #408, opened 2023-03-10
Move visually hidden space inside element

peteryates:
The final decision on how to properly handle the space between the visible and hidden text hasn't been made yet but the general consensus upstream is that it's better inside the visually hidden span than before it.

This might be revisited again when a decision has been made and the solution presented in #403 can be used instead.

### DIFF
--- a/app/components/govuk_component/summary_list_component/action_component.rb
+++ b/app/components/govuk_component/summary_list_component/action_component.rb
@@ -20,7 +20,7 @@ class GovukComponent::SummaryListComponent::ActionComponent < GovukComponent::Ba
 
   def call
     link_to(href, **html_attributes) do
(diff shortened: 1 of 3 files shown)


frankieroberto (approved):
👍 I think we can merge this straight away, as it’s what the [nunjucks component already does](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/summary-list/template.njk#L5)?

Do we consider this a breaking change? I guess it might break some downstream tests, depending on how they were written?

peteryates:
Yeah it's definitely a breaking change but the next release will be a big one anyway so the timing is good.

peteryates:
I added a note to the release notes issue #390

Issue #409, opened 2023-03-21
Accordion section IDs should be derived using the component ID rather than headings


edwardhorsford:
See related conversation on [cross gov slack](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1679416057509069).

The accordion component seems to [derive the section IDs](https://github.com/DFE-Digital/govuk-components/blob/main/app/components/govuk_component/accordion_component/section_component.rb#L18) using the heading for the section. This is different than how the design system does it, where they are [derived using the component id and an incrementing count](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/accordion/template.njk#L18).

## Description of bug
Unfortunately this causes a bug with the javascript of the accordion that attempts to preserve the open status of individual sections. The bug will mean that unrelated accordions will end up sharing their section open status.

The javascript attempts to keep preserve open status of sections across page loads. The docs for the component suggest this is done using the ID of the component. Teams are meant to give accordions unique IDs (across the domain, not the page), and the js will use this to persistently save the open status. Under the hood, I think the js actually uses the IDs of the section, and not the IDs of the component. 

In the design system the IDs of the sections are derived from the component ID - so as long as the component ID is unique across the domain, so will the section IDs. However with the implementation in this repo, a unique ID on the component has no bearing on the section IDs. Repeated accordions on different pages with different IDs will still get the same section IDs. The result is that the 'open' status of each section will be incorrectly shared between them.

## Suggested solution

Change the section IDs to be derived based on the component ID, where set.

peteryates:
I believe #410 resolves this in the best way possible. Setting an id on the accordion itself now propagates down to the sections, which should make it easier to enforce uniqueness. If it's still not enough I can take a proper look at inserting sequential numbers too.

edwardhorsford:
Thanks @peteryates!

PR #410, opened 2023-03-21
Use the accordion id to prefix its section ids

peteryates:
The GOV.UK Design System accordion component saves the state of open/closed accordions sections in localStorage, but it only saves the section's ID. If there are components on different pages that share IDs, the accordion could render in the incorrect state (with the state of the _other_ matching accordion section on another page.)

While ensuring global uniqueness is beyond the scope of this gem, we can increase the chances of uniqueness by incorporating the outer accordion's ID into each section ID.

Refs #409

### DIFF
--- a/app/components/govuk_component/accordion_component.rb
+++ b/app/components/govuk_component/accordion_component.rb
@@ -7,16 +7,16 @@ class GovukComponent::AccordionComponent < GovukComponent::Base
       html_attributes: html_attributes,
       summary_text: summary_text,
       heading_text: heading_text,
-      accordion_id: id,
+      accordion_id: accordion_id,
       &block
     )
   end
 
-  attr_reader :id, :heading_level
+  attr_reader :accordion_id, :heading_level
 
   def initialize(heading_level: 2, classes: [], html_attributes: {})
     @heading_level = heading_tag(heading_level)
-    @id            = html_attributes[:id]
+    @accordion_id  = html_attributes[:id]
 
     super(classes: classes, html_attributes: html_attributes)
   end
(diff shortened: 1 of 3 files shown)


frankieroberto:
> I'm not against the idea but the bug is also present upstream in `govuk-frontend`. The implementation here is actually safer than [the original](https://github.com/alphagov/govuk-frontend/blob/main/package/govuk/components/accordion/accordion.js#L1890) because that appears to assign incrementing numeric ids. I think the only reason it doesn't exhibit itself on the Design System docs is because of the `<iframe>` safety net.

The logic behind using an incrementing numeric ID was that it'd be guaranteed to be unique within the component - and then you just had to set a unique ID for the overall component.

Using `heading_text.parameterize` is probably fine, but would fail if you had 2 headings in the accordion with the same text (or which got paramaterized to the same text), which is admittedly highly unlikely?

> I suspect they might tweak their implentation a bit so the identifier saved to localStorage concatenates the `window.location.pathname` along with the header text; that _should_ provide app-wide uniqueness.

This is definitely worth considering, at least as an option. One downside is that there might be some services where they'd want the state to be preserved across multiple separate pages (although I can't think why), or they might have a single page which can have multiple paths (eg with and without a trailing `/`).

peteryates:
The reason it doesn't use numeric is because by the time we're [looping through the sections in the template](https://github.com/DFE-Digital/govuk-components/blob/main/app/components/govuk_component/accordion_component.html.erb#L2) the section HTML has already been rendered. I 

If it allowed a hash to be passed in containing all the sections (like we do for tables), injecting an index would be easier, but because lots of accordion sections contain many elements (paragraphs, lists, tables, etc) it's blocks all the way down.

Now I'm more familiar with ViewComponent I suspect there's a way of exposing the index, but I don't think there's a legit situation where two accordion sections have the same header text.

peteryates:
Here's the issue proposing the addition of an `#index` method to slots.

https://github.com/ViewComponent/view_component/issues/1620

PR #413, opened 2023-04-18
Fix show_actions_column overwrite

vassyz:
If the `show_actions_column` option is set to `false`, it should not be overwritten when the rows have actions. This was an issue when generating rows dynamically and wanting to hide the actions column in certain cases.


peteryates (approved):
Great, thank you!

PR #414, opened 2023-04-21
Remove the disabled button param

peteryates:
The `govuk-button--disabled` class [is going to be removed from the design system](https://github.com/alphagov/govuk-frontend/pull/3187) and its use is discouraged. Links styled as buttons should not be disabled and buttons should be disabled using the `disabled` HTML attribute instead.

### DIFF
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -12,7 +12,6 @@ module GovukLinkHelper
   }.freeze
 
   BUTTON_STYLES = {
-    disabled:  "govuk-button--disabled",
     secondary: "govuk-button--secondary",
     warning:   "govuk-button--warning",
   }.freeze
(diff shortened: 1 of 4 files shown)


frankieroberto:
@peteryates is it worth us following the pattern of `govuk-frontend` and keeping the feature but marking it as deprecated until the next major version release? We could add something like this for link:

```ruby
logger.warn('Deprecated: Links that are styled to look like disabled buttons will not be 
supported by future releases. Avoid using disabled buttons where possible, 
or use govuk_button_to to generate a disabled button')
```

For the `govuk_button_to` helper, should we keep the `govuk_button_to(text,url, {disabled: true })` syntax, but have `disabled: true` add the `disabled` attribute instead of a CSS class? That way it'd be backwards-compatible?

peteryates:
Thinking about it I'm just going to leave it as-is for this release. In the next release `disabled: true` will just add the disabled attribute.

PR #419, opened 2023-05-29
New link helpers

peteryates:
This PR introduces a new set of link helpers that will ~be available as an option but will eventually become default~ be the default. The legacy 100% Rails-compatible helpers will still be available for the time being by including the helper module as described below.

The current helpers are closely tied to Rails' implementation which after 18 years is showing its age. Because it pre-dates keyword arguments and can take many formats of link, the code is complex and difficult to extend.

The new implementation:

* is clean, all of the logic is handled by Rails' [class_names](https://www.rubydoc.info/docs/rails/ActionView%2FHelpers%2FTagHelper:class_names) helper
* does minimal argument juggling
* supports [opening links in new tabs](https://design-system.service.gov.uk/styles/typography/#opening-links-in-a-new-tab)
* **does not support** the old-style verbose non-resource-orientated `govuk_link_to("Show profile", controller: "profiles", action: "show", id: @profile)` format. I haven't seen much evidence of this style in use and Rails' documentation [advises against using it](https://api.rubyonrails.org/v7.0/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to)

## Multiple versions of link helpers

For a while there will be two versions of the link helpers:

1. the new ones `GovukLinkHelper`
2. the old ones `GovukRailsCompatibileLinkHelper`

I originally planned to make choosing between them a config option but due to the order the gem is required and then the intializers run, it's a pain. Instead, the new ones will automatically be included and if the old ones are wanted they can be manually included in `ApplicationHelper`:

```ruby
module ApplicationHelper
  include GovukRailsCompatibileLinkHelper
end
```

This will make the old method variants override the new ones.

I will begin adding some warnings to the old ones before they are dropped.

## Final things to do

* [ ] add some tests ensuring additional arguments (both `options` and `html_options`) can still be passed through to the Rails helpers
* [ ] `visually_hidden_prefix` and `visually_hidden_suffix` keyword arguments?


frankieroberto:
@peteryates I tried to look at this diff, but there’s a lot of changes so it's quite hard to follow (at least for me).

Would you mind summarising the main changes to the helpers that this PR proposes? Maybe a before & after would help?

peteryates:
@frankieroberto yeah, good call. I added a replacement helper that has the same methods in it and I think the fact I changed the order has made the diff more difficult to digest than it should be. I'll see if I can improve it and write a full summary.

Essentially though the old helpers follow a format like this:

```ruby
  def govuk_link_to(name = nil, options = nil, extra_options = {}, &block)
    extra_options = options if block_given?
    html_options = build_html_options(extra_options)

    if block_given?
      link_to(name, html_options, &block)
    else
      link_to(name, options, html_options)
    end
  end
```

As you can see here, `options` and `extra_options` are both hashes that are 'juggled' depending on whether link text or a block is provided. This makes the code hard to follow, especially once you look in [build_html_options](https://github.com/DFE-Digital/govuk-components/blob/main/app/helpers/govuk_link_helper.rb#LL95C7-L95C25) which goes to great lengths to merge the provided options with the ones required by the design system.

It's very brittle and caught me out when I tried to add the `new_tab` functionality as a keyword arg #363  some time ago. @hetties and I tried the preview release with Register and it broke half the test suite because [different versions of Ruby treat keyword args and argument hashes differently](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) - so I retracted it entirely.

This change switches everything over to keyword arguments, allowing the code to be much simpler. Most projects shouldn't notice any change. It might bite some that are using the old style Rails format, but the Rails docs have advised against using them for years and I barely see any in the wild so I don't image it's widespread. They will be able to continue using the old ones for the time being.

frankieroberto:
@peteryates ah ok, so it’d still be the case that the first argument would be the link text and the second one the path/url, but after that you’d use keyword arguments for any extra attributes or options?

peteryates:
This is likely to be one of the features of v5.0.0 which will be released following govuk-frontend 5.

peteryates:
@frankieroberto - in terms of reviewing this, the diff will be confusing because the change moves the old link helpers to [a different file](https://github.com/x-govuk/govuk-components/blob/0147f3f73762cfcb4ae0ab8d7a6dc00748bfa798/app/helpers/govuk_rails_compatibile_link_helper.rb) and adds [new implementations in their place](https://github.com/x-govuk/govuk-components/blob/0147f3f73762cfcb4ae0ab8d7a6dc00748bfa798/app/helpers/govuk_link_helper.rb).

The method names are the same so if people want to use the originals they can manually include the module, but the new ones will be default. The code is _much cleaner_ in the new ones so they should be easy to review on their own.

In making this change we lose the old Rails way of creating links, but that's actively discouraged by the Rails docs and you don't really see it any more (although we do have some in ECF!)

frankieroberto:
@peteryates thanks, I'll try to review!  Would you mind updating the PR description to show what the old and new link syntax looks like? That might make it a bit clearer.

peteryates:
Have done, and here's a comparison of old vs new:
```ruby
# old
link_to("Show profile", controller: "profiles", action: "show", id: @profile)

# new (restful)
link_to("Show profile", profile_path(@profile))
```

frankieroberto:
> Have done, and here's a comparison of old vs new:

Thanks! I've not seen anyone use the old style for ages! Presumably both examples should be `govuk_link_to` not `link_to`?

Will anyone using the old style get an error, or will it just render a link but in an unexpected way?

peteryates:
> Presumably both examples should be govuk_link_to not link_to?

I was just showing the Rails defaults when demonstrating the formats, but I've chagnged it to `govuk_link_to` in the PR description 👍🏽 

> Will anyone using the old style get an error, or will it just render a link but in an unexpected way?

They'll get an `ArgumentError: unknown keyword: :controller` or similar.

app/helpers/govuk_link_helper.rb
```diff
@@ -3,156 +3,123 @@
 module GovukLinkHelper
   using HTMLAttributesUtils
 
-  def govuk_link_classes(*styles, default_class: "#{brand}-link")
-    if (invalid_styles = (styles - link_styles.keys)) && invalid_styles.any?
-      fail(ArgumentError, "invalid styles #{invalid_styles.to_sentence}. Valid styles are #{link_styles.keys.to_sentence}")
+  def govuk_link_classes(inverse: false, muted: false, no_underline: false, no_visited_state: false, text_colour: false)
+    if [text_colour, inverse, muted].count(true) > 1
+      fail("links can be either text_colour, inverse or muted - not combinations of the three")
```
frankieroberto, line 8:
```suggestion
      fail("links can only be one of text_colour, inverse or muted")
```

app/helpers/govuk_rails_compatibile_link_helper.rb
```diff
@@ -0,0 +1,157 @@
...
+    [default_class] + link_styles.values_at(*styles).compact
+  end
+
+  def govuk_button_classes(*styles, default_class: "#{brand}-button")
+    if (invalid_styles = (styles - button_styles.keys)) && invalid_styles.any?
+      fail(ArgumentError, "invalid styles #{invalid_styles.to_sentence}. Valid styles are #{button_styles.keys.to_sentence}")
+    end
+
+    [default_class] + button_styles.values_at(*styles).compact
+  end
```
frankieroberto, line 20: Should these be private methods? Can't think of why you'd use them outside of a link or button helper.

If not, maybe move them below the other methods, and add a brief comment to explain what they do? It threw me a bit seeing these alongside the more obvious helper methods below.

app/helpers/govuk_rails_compatibile_link_helper.rb
```diff
@@ -0,0 +1,157 @@
...
+
+  def govuk_link_to(name = nil, options = nil, extra_options = {}, &block)
+    extra_options = options if block_given?
+    html_options = build_html_options(extra_options)
+
+    if block_given?
+      link_to(name, html_options, &block)
+    else
+      link_to(name, options, html_options)
+    end
```
frankieroberto, line 30: I think you can avoid the `if` conditions here and always pass `block` on to the rails helpers, as those do their own `if block_given?` checks?

```suggestion
    extra_options = options if block_given?
    html_options = build_html_options(extra_options) || nil

    link_to(name, options, html_options, &block)
```

guide/lib/helpers/class_helpers.rb
```diff
@@ -0,0 +1,7 @@
+module Helpers
+  module ClassHelpers
+    def class_names(*args, **kwargs)
```
frankieroberto, line 3: Make this a private method where it's needed instead? Quite hard to interpret what this does when looking at it out of context!

spec/helpers/govuk_rails_compatibile_link_helper_spec.rb
```diff
@@ -0,0 +1,325 @@
...
+          expect(subject).to match_array(['govuk-button', css_class])
+        end
+      end
+    end
+  end
+
+  describe "#govuk_link_to" do
+    let(:link_text) { 'Onwards!' }
+    let(:link_url) { '/some/link' }
+    let(:link_params) { { controller: :some_controller, action: :some_action } }
```
frankieroberto, line 71: Confused by this - thought you’d deprecated this method, and now always require a path to be set instead of setting a controller and action?

frankieroberto (approved):
Left some comments/suggestions - but feel free to ignore. Looks good!

guide/lib/helpers/class_helpers.rb
peteryates: This is only used within the guide, it's a crude reimplementation of Rails' `#class_names` - now the guide pulls in more of Rails, it might be possible to remove this altogether.

spec/helpers/govuk_rails_compatibile_link_helper_spec.rb
peteryates: This covers the old helpers which are still present (in `GovukRailsCompatibileLinkHelper`) for teams working on codebases that won't support the new style.

guide/lib/helpers/class_helpers.rb
peteryates: Yeah, rebasing from `main` means this is no longer needed! Thanks for highlighting.

app/helpers/govuk_rails_compatibile_link_helper.rb
peteryates: So these are exposed to make it [easier for devs to use Rails' _other_ link helpers](https://govuk-components.netlify.app/helpers/link/#class-helpers), like `link_to_if` and `link_to_unless`, `link_to_unless_current`, etc.

app/helpers/govuk_rails_compatibile_link_helper.rb
peteryates: Going to leave these as they're in the legacy implementation which will soon be dropped.

PR #426, opened 2023-06-22
Use new slots syntax in docs

joelhawksley:
This PR updates a docs page to use the new Slots syntax ❤️

### DIFF
--- a/guide/content/introduction/about-viewcomponent.slim
+++ b/guide/content/introduction/about-viewcomponent.slim
@@ -42,9 +42,9 @@ markdown:
 
         / with slots:
         = render(ShoppingList.new) do |shopping_list|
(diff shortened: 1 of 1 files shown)


frankieroberto (approved):
Thank you! This got missed.

Incidentally I still don’t completely understand the rationale behind the new `with_` prefix. Is it to avoid method name collisions?

joelhawksley:
> Incidentally I still don’t completely understand the rationale behind the new with_ prefix. Is it to avoid method name collisions?

In a way, yes. Previously the getter and setter were the same method, restricting our ability to expand the slots feature set.

peteryates:
Totally missed that, thank you @joelhawksley 🌟

PR #430, opened 2023-06-25
Align documentation site with other X-GOVUK projects

paulrobertlloyd:
Possibly some controversial suggestions in this PR, but all up for debate (not just for this project, but others linked from the X-GOVUK website, too).

### 1. Licence and copyright

↳ Fixed in #431

* Removed Open Government Licence as not relevant or applicable in this context, and instead link to MIT Licence in repository.
* Added copyright notice against X-GOVUK.

Only changed in documentation footer, but wondering if corresponding changes need to be made in the `LICENCE` file also. Is this project covered by Crown Copyright? Is copyright held by DfE? Can it be transferred by updating a text file!? Unsure.

### 2. Project name

↳ Fixed in #432

* Added X-GOVUK logo
* Changed GOV.UK to X-GOVUK
* Changed name to Components for Ruby on Rails

What is the correct, canonical name for this project? It seems to vary. Separately, been struggling with many of our projects being called ‘GOV.UK [Project name]’ but appearing on documentation site headers as ‘X-GOVUK [Project name]’. Not quite sure how to resolve that tension.

### 3. Typeface

* Changed to using `system-ui, sans-serif`

This means loading less assets for one, but also looks different enough from GOV.UK to prevent confusion.

I know you purchased the font to get around GDS Transport licensing and usage restrictions, but its not _quite_ the same, and as mentioned, it currently renders incorrectly in Safari, though now fixed in Safari v17.

`system-ui` looks great on a Mac (uses SF Text), but honestly not as nice on Windows 10 from my experience (though nothing looks good there) and not sure which fonts will get used on different Linux distros.

Thought it might be possible to use Transport for examples only, but given the way GOV.UK Frontend works, almost impossible to override, or at least not easily.

### 4. Breadcrumbs

* Added breadcrumb in the home page masthead linking back to the main X-GOVUK website.

This is modelled on the GOV.UK SaaS services which do similar by linking back to a landing page on the [GOV.UK Service Manual](https://www.gov.uk/service-toolkit#gov-uk-services).

FWIW, this link is broken as the `shared-projects` anchor is missing on the X-GOVUK website since the heading was changed to ‘Projects’. Maybe the link should say ‘X-GOVUK Projects’?

Also added this to interior pages… though I’m kinda wondering if breadcrumbs are needed here? Also, while the GOV.UK SaaS websites include the page title in the breadcrumb, that’s not usually how breadcrumbs are used across GOV.UK.

***

[...]

PR #431, opened 2023-06-26
Update licence

paulrobertlloyd:
(Or how being British while using a technology that uses International English can be painful sometimes)

* Renamed LICENCE to LICENSE.txt to be consistent with other X-GOVUK projects and reflect the fact that the MIT License uses ~~American~~ International English spelling for ‘licence’
* Use ‘licence’ in places where UK English is (and can) used
* Add X-GOVUK to LICENSE.txt
* Update footer in documentation site to remove OGL and link to MIT License, and add X-GOVUK copyright statement

See discussion in #430.


paulrobertlloyd:
Tests failing because linking to a renamed LICENSE.txt file, which will exist once this PR is merged. Kinda like inception…

paulrobertlloyd:
> Having an inconsistency between the filename and the content within feels a bit odd

To be clear:

* Before: `LICENCE`
* After: `LICENSE.txt`

The ‘licen **c** e’ (Brit.) for this project is the ‘MIT Licen **s** e’ (Intl.), so updating the file name does match the title of the licence. 🤪

peteryates (approved):

PR #437, opened 2023-07-08
Add the new exit this page component

peteryates:
The [Exit this page component](https://design-system.service.gov.uk/components/exit-this-page/) allows users who are at risk to immedaitely and safely leave a page.

* [x] Add a guide page
* [x] Add a component helper
* [x] Add specs that cover the configuration options
* [x] Add entry to the README


app/components/govuk_component/exit_this_page_component.rb
```diff
@@ -0,0 +1,52 @@
+class GovukComponent::ExitThisPageComponent < GovukComponent::Base
+  attr_reader :text, :redirect_url, :secondary
+
+  def initialize(text: config.default_exit_this_page_text, redirect_url: config.default_exit_this_page_redirect_url, secondary: false, classes: [], html_attributes: {})
+    @text = text
+    @redirect_url = redirect_url
+    @secondary = secondary
```
peteryates, line 7: I'm not really sure of the best way of naming this. In the design system the section is called ['adding the secondary link'](https://design-system.service.gov.uk/components/exit-this-page/).

Secondary isn't very descriptive though. Would it be better if it was called `skip`? `skip_link`?

paulrobertlloyd: I think `skip_link`, given that’s what’s used in the code, and the intention of this feature.

Meanwhile, this is annoying: the 2 classes used are inconsistently named:

* `govuk-skip-link`
* `govuk-js-exit-this-page-skiplink`

😵‍💫 I wonder if there’s a naming convention being applied here? Anyway, nothing to do with the project.

peteryates: Yeah, I wondered about just adding an option to the Skip Link helper but that didn't feel quite right. I'll change it to `skip_link`.

frankieroberto: I'm not sure it makes sense to have the secondary link be part of the same component. Feels to me that it'd be more intuitive to add the option to the existing skip link component? That’s what `govuk-frontend` does too?

paulrobertlloyd: Ah yes, I see how this works. Yup, I agree with @frankieroberto.

peteryates: That felt unintuitive to me, I'd expect the exit this page functionality to all be in one place.

paulrobertlloyd: No; the exit this page button appears in a certain location before the page content, where as this second skip link appears after the other skip link, before the header and navigation. So these components would need to be placed independently within a layout.

paulrobertlloyd: Given these components need to live in separate parts of the page, I’m not sure how else you could have implemented it?

peteryates: I meant from a code organisation point of view, I'd expect the code that generates 'exit this page' HTML to be accessible via `govuk_exit_this_page`, not `govuk_skip_link`.

The upstream implementation of this doesn't feel right:

```
{% block skipLink %}
  {{ super() }}
  {{ govukSkipLink({
    href: "https://bbc.co.uk/weather/",
    classes: "govuk-js-exit-this-page-skiplink"
  }) }}
{% endblock %}
```

Am I skipping to BBC Weather?

I get that it's piggybacking on another component but it feels clunky and less well thought out than the rest of the design system.

peteryates: I prefer the second option.

Also, I'm not hugely keen on the component having a `redirect_url` and the link having a `href`. I might just let the component take either.

peteryates:
Looking very decent now, thanks @paulrobertlloyd and @frankieroberto 

One weird thing is that on the [example for the secondary link](https://deploy-preview-437--govuk-components.netlify.app/components/exit-this-page/#adding-a-secondary-link-to-the-top-of-the-page), hitting tab then enter takes me to Wikipedia even though it's a link to BBC Weather.

The JavaScript must expect the component to be used once per page.

paulrobertlloyd:
> The JavaScript must expect the component to be used once per page.

Might be worth filing this issue upstream; while it should only be used once, perhaps an indication that the underlying JavaScript could be more robust.

paulrobertlloyd:
If you look at the Nunjucks component, the HTML is intended to be used *within* the button (i.e. you can have a button with strong text). [This component is basically a button that lives inside a `<div>`](https://github.com/alphagov/govuk-frontend/blob/support/4.x/src/govuk/components/exit-this-page/template.njk).

Can `govuk_button_link_to` accept HTML content?

paulrobertlloyd:
Looking at the component code:

```rb
def exit_this_page_content
  content.presence || govuk_button_link_to(text, redirect_url, **link_attributes)
end
```

This should probably be something like (but most likely not) this:

```rb
def exit_this_page_content
  govuk_button_link_to(text || content.presence, redirect_url, **link_attributes)
end
```

peteryates:
You were pretty much bang on 😅  [Amended here] and updated the guide example accordingly (strong text on the inside of the button seems like a fair example).

paulrobertlloyd:
> You were pretty much bang on

My powers in Ruby must be growing in strength 🦹‍♂️

paulrobertlloyd:
Taking of supervillains, being super nit-picky, but can/should the ERB example be this:

```erb
<%= govuk_exit_this_page(redirect_url: "https://www.wikipedia.org") do %>
  <strong>Exit this page now</strong>
<% end %>
```

(i.e. content on a new line. Would this introduce white space issues?)

peteryates:
I can definitely take a look at making the ERB output nicer, but that'd be a separate change. You can take a look at the steps involved to get it to where it is [here](https://github.com/x-govuk/govuk-components/blob/main/guide/lib/helpers/formatters.rb#L44) - warning, it's not the prettiest code I ever wrote.

app/helpers/govuk_exit_this_page_link_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+    **html_attributes,
+    &block
+  )
+    link_classes = Array.wrap(classes).append(%w(govuk-skip-link govuk-js-exit-this-page-skiplink))
+
+    html_attributes_with_data_module = { data: { module: "govuk-skip-link" } }.deep_merge(html_attributes)
+
+    return link_to(href, class: link_classes, **html_attributes_with_data_module, &block) if block_given?
+
+    link_to(text, href, class: link_classes, **html_attributes_with_data_module)
```
frankieroberto, line 15: Would this work - passing both `block` and `text` to `link_to` and relying on [the logic there](https://github.com/rails/rails/blob/main/actionview/lib/action_view/helpers/url_helper.rb#L212-L222) to have the block override the text param?

```suggestion
    link_to(text, href, class: link_classes, **html_attributes_with_data_module, &block)
```

frankieroberto (approved):
Looks good to me!

app/helpers/govuk_exit_this_page_link_helper.rb
peteryates: Good point, that works just fine.

peteryates:
Added one more minor(ish) change - now the assistive text that can be configured in the Nunjucks macros can be configured here too via the keywords:

* `activated_text`
* `timed_out_text`
* `press_two_more_times_text`
* `press_one_more_time_text`

And of course these can be configured app-wide in the usual fashion. This is now done unless there are any bugs or things missing.

paulrobertlloyd (approved):

Issue #440, opened 2023-07-13
Support other design systems


tvararu:
Like https://github.com/x-govuk/govuk-form-builder/pull/127, is there any appetite to support the `config.brand` option in this library as well?

We're using [the form builder option in a `nhsuk-frontend` project](https://github.com/nhsuk/record-childrens-vaccinations/pull/327), and while we had to patch over the JS for the radio buttons, the rest of it seems to work. Having that option was a big help.

If it sounds good, maybe I could make a start on it.

peteryates:
Yeah absolutely - I'd be more than happy to have that functionality here too.

PR #441, opened 2023-07-14
Support other design systems

tvararu:
This is mostly identical to the work done in https://github.com/x-govuk/govuk-form-builder/pull/127.

Commits kept separate for each individual component in case it makes it easier to review, but should be squashed up before merge.

Fixes https://github.com/x-govuk/govuk-components/issues/440.

### DIFF
--- a/app/components/govuk_component/accordion_component.rb
+++ b/app/components/govuk_component/accordion_component.rb
@@ -24,7 +24,7 @@ class GovukComponent::AccordionComponent < GovukComponent::Base
 private
 
   def default_attributes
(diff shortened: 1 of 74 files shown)


tvararu:
Converting this to draft as I'm just testing it locally and it seems like the links don't appear correctly.

peteryates:
I've added you as a collaborator, if you re-raise the PR from a branch the guide will build.

Changes look good by the way, really looks like it's taking shape.

tvararu:
Superseded by https://github.com/x-govuk/govuk-components/pull/442.

PR #442, opened 2023-07-17
Support other design systems

tvararu:
This is https://github.com/x-govuk/govuk-components/pull/441 but re-raised from a branch on this repo.

This is mostly identical to the work done in https://github.com/x-govuk/govuk-form-builder/pull/127.

Commits kept separate for each individual component in case it makes it easier to review, but should be squashed up before merge.

Fixes https://github.com/x-govuk/govuk-components/issues/440.

TODO:

- [x] Fix links
- [x] Squash superfluous commits

### DIFF
--- a/app/components/govuk_component/header_component.rb
+++ b/app/components/govuk_component/header_component.rb
@@ -44,22 +44,22 @@ class GovukComponent::HeaderComponent < GovukComponent::Base
 private
 
   def default_attributes
-    { class: %w(govuk-header) }
+    { class: ["#{brand}-header"] }
   end
 
   def navigation_html_attributes
-    nc = %w(govuk-header__navigation).append(custom_navigation_classes).compact
+    nc = ["#{brand}-header__navigation"].append(custom_navigation_classes).compact
 
     { class: nc, aria: { label: navigation_label } }
   end
 
   def container_html_attributes
-    { class: %w(govuk-header__container govuk-width-container).append(custom_container_classes).compact }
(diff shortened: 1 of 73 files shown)


tvararu:
This seems to be working now for our app. PR is ready for review.

app/components/govuk_component/header_component.rb
```diff
@@ -44,22 +44,22 @@ def initialize(classes: [],
 private
 
   def default_attributes
-    { class: %w(govuk-header) }
+    { class: ["#{ brand }-header"] }
```
peteryates, line 47: I haven't seen the spacing within an interpolated string style before. I can see how it improves readability, especially on bigger blocks of text. We now have a mix of with and without though.

guide/content/introduction/configuration.slim
```diff
@@ -15,6 +15,7 @@ markdown:
 pre
   code.highlight.language-ruby
     | Govuk::Components.configure do |conf|
+        conf.brand = "swanky-new-design-system" # or "govuk"
```
misaka, line 18: If this is where people are likely to first encounter this configuration point, it may be worth making it clear the default is "govuk":

```suggestion
        conf.brand = "swanky-new-design-system" # default is "govuk"
```

tvararu:
Comments addressed and commits squashed into one

peteryates (approved):
Magnificent ✅ 

Great work, thank you @tvararu ❤️

peteryates:
It would be nice as a follow up to include a section in the guide showing off an NHS styled component. We can use the `rendered_partial` argument so we can render it in an iframe and not have to load two sets of CSS.

PR #445, opened 2023-08-23
Add task list component

peteryates:
This PR adds the upcoming [Task list component](/alphagov/govuk-design-system/pull/1994). It's built in a similar fashion to the summary list.

Tags can be added within statuses by calling `govuk_tag` within the status, or plain text can be used.

Screenshot 2023-10-31 at 10 30 53

## Final tasks

* [ ] Add the ability to initialise the component with a Nunjucks style array of hashes.
* [x] Finish off the guide page
* [x] Remove the temporary task list component styles


guide/content/stylesheets/application.scss
```diff
@@ -104,3 +104,77 @@ $govuk-inverse-button-shadow-colour: govuk-shade($govuk-inverse-button-text-colo
     }
   }
 }
+
+@include govuk-exports("govuk/component/task-list") {
```
frankieroberto, line 108: Is all this CSS just temporary, so that this can be tested before the task list component gets shipped with govuk-frontend?

app/components/govuk_component/task_list_component/status_component.rb
```diff
@@ -0,0 +1,25 @@
...
+      @href = href
+      @colour = colour
+
+      super(classes: classes, html_attributes: html_attributes)
+    end
+
+    def call
+      tag.div(**html_attributes) do
+        render(GovukComponent::TagComponent.new(text: text, colour: colour))
+      end
```
frankieroberto, line 16: Hmm, so one tricky thing with the new task list component is that using the tag component within the status is 'optional', and there also needs to be a way to just pass in plain text or html. This is recommended for 'Completed' items, where there’s no need for a coloured tag, and instead just the plain text of "Completed" can be used.

In the Nunjucks macro, this is done by having an optional `tag` object within the `status` object, like this:

```javascript
items: [
  {
    title:
      {
        text: "Company Directors"
      },
    href: "#",
    status: {
      text: "Completed"
    }
  },
  {
    title:
      {
        text: "Registered company details"
      },
    href: "#",
    status: {
      tag: {
        text: "Incomplete",
        classes: "govuk-tag--blue"
      }
    }
  }
]
```

for the Ruby component, maybe we could do something similar, like this:

```ruby
<%= govuk_task_list do |task_list|
  task_list.with_item(title: "Contact details", href: '#', status: "Completed")
  task_list.with_item(title: "Project details", href: '#', status: {tag: {text: "Incomplete", colour: "green"}})
end %>
```

Or alternatively we don't do anything smart with the arguments, and instead get people to use the tag helper directly:

```ruby
<%= govuk_task_list do |task_list|
  task_list.with_item(title: "Contact details", href: '#', status: "Completed")
  task_list.with_item(title: "Project details", href: '#', status: govuk_tag(text: "Incomplete", colour: "green"))
end %>
```

...which might be simpler?

app/components/govuk_component/task_list_component/item_component.rb
```diff
@@ -0,0 +1,45 @@
...
+      when String
+        with_status(text: status)
+      when Hash
+        with_status(**status)
+      else status
+      end
+    end
+
+    def hint_content
+      tag.div(hint, class: %w(govuk-task-list__task_hint))
```
frankieroberto, line 38: The hint is a bit more complicated, as it needs an `id` attribute setting, and then that ID should get added to the `aria-describedby` attribute on the link (space separated before the ID for the status div).

See [nunjucks code](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/task-list/template.njk#L11) for reference.

app/components/govuk_component/task_list_component/title_component.rb
```diff
@@ -0,0 +1,37 @@
...
+  private
+
+    def title_content
+      return link if href.present?
+
+      text
+    end
+
+    def link
+      govuk_link_to(text, href, class: "govuk-task-list__link")
```
frankieroberto, line 26: This needs an `aria-describedby` attribute which references the status (and the hint if there is one).

frankieroberto (changes requested):
Looking good! Left some initial comments.

guide/content/stylesheets/application.scss
peteryates: Yeah, I [left a reminder](https://github.com/x-govuk/govuk-components/pull/445/commits/5872342e0a3cc5f22bb595593c38093f7a54c781) - I added an item to the checklist above so we don't forget.

app/components/govuk_component/task_list_component/status_component.rb
peteryates: Ah, I didn't realise regular text was acceptable too - I like the second option.

peteryates: I had a play with this and passing in raw HTML like `status: govuk_tag(...)` and we need to ensure it's safe, which here adds some risk.

I think this needs to be done via the block:

```erb
<%= govuk_task_list do |task_list|
  task_list.with_item(title: "Project details", href: '#') do |item|
    item.with_title(text: "...", href: "...", hint: "...")
    item.with_status { govuk_tag(text: "Incomplete", colour: "green")) }
  end
end %>
```

frankieroberto: You’ve got 2 titles and hrefs there…

I forget exactly how the html_safe system works in Rails, but I do think the  params version is easier to use than the block one, if we can figure it out.

peteryates: Oops, I forgot to remove the args, I've updated the snippet.

The problem here is that we need to call `#html_safe` on the `status` param, but that could contain anything, like if a team made dynamic statuses for example.

guide/content/stylesheets/application.scss
frankieroberto: @peteryates I've merged the v5 branch into this branch and have removed this css.

peteryates:
I've added support for `aria-descrribedby` in 2bf5a18

There are some slight complications:

* in the examples the `aria-describedby` attribute is on the `<a>` element but we now support rendering without the link. When that happens I've moved the `aria-describedby` up to the `div.govuk-task-list__name-and-hint`.
* there's no nice way of us talking to sibling components with ViewComponent, so the `Title` and `Status` both belong to `Item` - it's impossible to know whether or not there's an status present when rendering the title. In Nunjucks where everything's passed in in a single object at the start it's easlier, but here less so. For the time being I'm assuming there's always a status.

frankieroberto:
@peteryates:
> * in the examples the `aria-describedby` attribute is on the `<a>` element but we now support rendering without the link. When that happens I've moved the `aria-describedby` up to the `div.govuk-task-list__name-and-hint`.

As I understand it, `aria-describedby` can only be added to interactive elements, so in the nunjucks macro we don’t add this attribute at all where there is no link – see [example](https://release-5-0--govuk-design-system-preview.netlify.app/patterns/complete-multiple-tasks/#tasks-that-cannot-yet-be-started).

frankieroberto (approved):

frankieroberto:
@peteryates do we think there’s anything left to do on this or shall we merge it into the v5 branch? Do we still want to support an array of item objects as well as the block syntax, or is that an optional extra?

peteryates:
I do plan on adding the array of items approach but I'm happy for that to follow as an enhancement, will get this merged later today.

PR #447, opened 2023-09-01
Add helper: title_with_error_prefix

frankieroberto:
This is simple helper for adding the "Error: " prefix to a page title depending on whether there are errors present.

This is required by the [GOV.UK Design System Error Summary component](https://design-system.service.gov.uk/components/error-summary/), so screen readers read it out as soon as possible.

Extracted from [the implementation in Apply for teacher training](https://github.com/DFE-Digital/apply-for-teacher-training/blob/596130f74dcb94ef7e03b7140c5b141c2375b999/app/helpers/view_helper.rb#L48-L50).

### DIFF
--- a/lib/govuk/components/engine.rb
+++ b/lib/govuk/components/engine.rb
@@ -38,6 +38,7 @@ module Govuk
     # +:default_breadcrumbs_hide_in_print+ false
     # +:default_cookie_banner_aria_label+ "Cookie banner"
     # +:default_cookie_banner_hide_in_print+ true
+    # +:default_error_prefix+ Text to use at the start of the page title tag when there is an error on the page. Default is 'Error: '
     # +:default_exit_this_page_redirect_url+ The URL that the exit this page component links to by default. Defaults to https://www.bbc.co.uk/weather
     # +:default_exit_this_page_text+ The default text that forms the link. Defaults to 'Exit this page'
     # +:default_exit_this_page_activated_text+ Text announced by screen readers when Exit this Page has been activated via the keyboard shortcut. Default in govuk-frontend is 'Exiting page.' Defaults to nil so govuk-frontend value is used unless overridden.
@@ -79,6 +80,7 @@ module Govuk
       default_breadcrumbs_hide_in_print: false,
       default_cookie_banner_aria_label: "Cookie banner",
       default_cookie_banner_hide_in_print: true,
+      default_error_prefix: "Error: ",
       default_exit_this_page_redirect_url: "https://www.bbc.co.uk/weather",
       default_exit_this_page_text: "Exit this page",
       default_exit_this_page_activated_text: nil,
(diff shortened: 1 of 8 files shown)


frankieroberto:
I did a quick search (thanks GitHub) and the same helper method is currently used on a number of DfE services:

* [Publish teacher training courses](https://github.com/DFE-Digital/publish-teacher-training/blob/a4592a2501db70d4276577e13888fbb9f57e99e0/app/helpers/view_helper.rb#L43-L45)
* [Get help with tech](https://github.com/DFE-Digital/get-help-with-tech/blob/731cf00974572b6d9123de59d6c72d72c2e116d9/app/helpers/view_helper.rb#L43-L45)
* [Data allowances prototype](https://github.com/DFE-Digital/data-allowances-prototype/blob/c114b08e767a6219aae4ff716ed36f0e2e9807f0/app/helpers/view_helper.rb#L16-L18)

Teaching Vacancies has a [more complicated helper](https://github.com/DFE-Digital/teaching-vacancies/blob/main/app/helpers/vacancies_helper.rb#L12-L18) which accepts some form of wizard step and a form object in order to build the title.

Get into Teaching has an alternative approach, where an [ErrorTitle is defined a middleware](https://github.com/search?q=repo%3ADFE-Digital%2Fget-into-teaching-app%20ErrorTitle&type=code) and searches the page for `.govuk-error-summary` and then adds the error prefix automatically if found. Interesting - if maybe a bit unconventional?!

No doubt there are some other approaches too! 🤔

lib/govuk/components/engine.rb
```diff
@@ -113,6 +113,7 @@ def reset!
 
       default_link_new_tab_text: "(opens in new tab)",
 
+      error_prefix: "Error: ",
```
peteryates, line 116: Ideally there'd be a line in the documentation block above explaining what this does.

frankieroberto: @peteryates done. I've also renamed it to `default_error_prefix` for consistency.

There’s currently not any way to override it within the helper itself though. Wonder if there should be? Maybe that's something to consider down the line if/when multiple languages are supported?

peteryates: Elsewhere I've set values like this in the following manner which is concise and seems to have worked well.

```ruby
  def title_with_error_prefix(title, error:, error_prefix: Govuk::Components.config.default_error_prefix)
    ...

```

frankieroberto: Ah, nice. Done in 574de503d5024c06cc0a1911641a0c7ee9b4a24f

peteryates (approved):
Fantastic addition 🚀

frankieroberto:
@peteryates I've added some quick documentation to the guide in 290893c57adca78827b46ea1d742e9cca9fb97b1 - let me know what you think!

What's the process for merging / releasing this? Guess we don't want it to appear in the documentation until it's been released..?

PR #449, opened 2023-09-25
Update govuk_button_to

frankieroberto:
This updates the behaviour of the `govuk_button_to` helper, so that:

* it adds the `data-module="govuk-button"` attribute to the button
* it always outputs a `<button>` element instead of `<input type="submit">` (this is now default behaviour in Rails 7)
* if `prevent_double_click: true` is set, `data-prevent-double-click="true"` is added to allow double click prevention (in common with the [`govuk_submit` helper on the formbuilder](https://govuk-form-builder.netlify.app/form-elements/submit))

### DIFF
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -43,7 +43,16 @@ module GovukLinkHelper
 
   def govuk_button_to(name = nil, options = nil, extra_options = {}, &block)
     extra_options = options if block_given?
-    html_options = build_html_options(extra_options, style: :button)
+    html_options = {
+      data: {module: "govuk-button"}
+    }
+
+    if extra_options && extra_options[:prevent_double_click]
+      html_options[:data]["prevent-double-click"] = "true"
+      extra_options = extra_options.except(:prevent_double_click)
+    end
+
+    html_options.merge! build_html_options(extra_options, style: :button)
 
     if block_given?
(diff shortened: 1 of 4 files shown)


frankieroberto:
@peteryates hmm so weirdly the test pass for this (asserting that it's using the Rails 7 default of always outputting a `<button>` tag), but the preview of the docs still uses the old default of `<input type="submit">` and I have no idea why.

Something to do with how the examples are rendered in the docs I guess?

peteryates:
Hmm interesting, I'll take a look tomorrow. The guide pulls in enough Rails to render things but it's not 1:1. Perhaps it's a config thing.

app/helpers/govuk_link_helper.rb
```diff
@@ -43,7 +43,16 @@ def govuk_mail_to(email_address, name = nil, extra_options = {}, &block)
 
   def govuk_button_to(name = nil, options = nil, extra_options = {}, &block)
     extra_options = options if block_given?
-    html_options = build_html_options(extra_options, style: :button)
+    html_options = {
+      data: {module: "govuk-button"}
```
peteryates, line 47:
```suggestion
      data: { module: "govuk-button" }
```

Rubocop should have caught this but it runs via `make ruby-lint` and predates this library's use of the `app` directory. I've updated that in #450

PR #455, opened 2023-10-23
Refactor `phase_tag_component` to fix failing spec

peteryates:
There's an unnecessary #deep_merge here that somehow seems to confuse the tests, the outer hash sees the overridden #brand but the inner one doesn't and reverts to "govuk", causing the 'ensure everything's overridden' test to fail.

I don't think this would cause any problems in a live environment where the brand value is set once during initialisation and left.


frankieroberto (approved):
Seems fine to me.

PR #456, opened 2023-10-26
Update to use govuk-frontend v5

frankieroberto:


peteryates:
I suggest we make this _the_ v5.0 branch and merge the other v5.0 work into it.

app/components/govuk_component/tag_component.rb
```diff
@@ -1,7 +1,7 @@
 class GovukComponent::TagComponent < GovukComponent::Base
   attr_reader :text, :colour
 
-  COLOURS = %w(grey green turquoise blue red purple pink orange yellow).freeze
+  COLOURS = %w(grey green turquoise blue light-blue red purple pink orange yellow).freeze
```
peteryates, line 4: Nice, missed this!

frankieroberto:
@peteryates I think this is now ready for us to distribute a pre-releases so that we and others can start testing it on some of our services?

app/components/govuk_component/header_component.html.erb
```diff
@@ -2,27 +2,22 @@
...
+        <% else %>
+          <svg
+            focusable="false"
+            role="img"
+            class="<%= brand %>-header__logotype"
+            xmlns="http://www.w3.org/2000/svg"
+            viewBox="0 0 152 30"
+            height="30"
+            width="152"
+          >
```
colinrotherham, line 16: Tiny little suggestion, see https://github.com/x-govuk/govuk-components/pull/466#issuecomment-1829691211 for more info

```suggestion
            width="152"
            aria-label="GOV.UK"
          >
```

colinrotherham: Ah sorry @frankieroberto, would you mind moving the `aria-label` in https://github.com/x-govuk/govuk-components/pull/456/commits/c96c555b925300eb375813fd7521476b20e7a0d9?

It needs to be on the `<svg>` element rather than the `<title>`

frankieroberto: Fixed in #468

PR #461, opened 2023-11-14
Fix capitalisation of viewBox

### LINKED ISSUE #459: SVG attribute `viewBox` not correctly capitalised
### Description of issue

While swapping out some hardcoded Design System HTML for the (excellent) view components in this gem, I noticed a small regression; in the hardcoded HTML the crown SVG has an attribute `viewBox`, while in the ERB provided in the gem it is called `viewbox`:

https://github.com/x-govuk/govuk-components/blob/30340f72ecb42086fc8ae641c3abef1a751f4414/app/components/govuk_component/header_component.html.erb#L11

https://github.com/x-govuk/govuk-components/blob/30340f72ecb42086fc8ae641c3abef1a751f4414/app/components/govuk_component/footer_component.html.erb#L29

According to Google SVG attributes are case sensitive, and this one should be kebab case (for references, [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox), this [tweet about the issue with replies from frontend experts](https://twitter.com/rem/status/759744091538137091?s=20), and this [article stating that SVG is case-sensitive](https://thenewcode.com/1151/Understanding-the-SVG-viewBox#:~:text=Also%20note%20the%20capitalization%20of,it%20is%20to%20be%20recognized.&text=You%20might%20think%20of%20this,the%20elements%20inside%20an%20SVG.)).

I'm not sure if this is actually an issue in any browsers; I can't see a difference in Chrome; but since the attribute is correctly capitalised [in the latest release of GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/v4.7.0/package/govuk/components/header/template.njk#L27), I thought I should flag it up.

### Suggested fix

Replace `viewbox` with `viewBox` in SVG tags.

frankieroberto:
Fixes #459

Thanks @lfdebrux for reporting this!


lfdebrux (approved):
Thanks for the speedy response!

app/components/govuk_component/footer_component.html.erb
```diff
@@ -26,7 +26,7 @@
           <% end %>
 
           <% if meta_licence.nil? %>
-            <svg aria-hidden="true" focusable="false" class="<%= brand %>-footer__licence-logo" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 483.2 195.7" height="17" width="41">
+            <svg aria-hidden="true" focusable="false" class="<%= brand %>-footer__licence-logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 483.2 195.7" height="17" width="41">
```
aliuk2012, line 29: @frankieroberto it might be worth adding a test for this because I can see someone coming along and reverting this change thinking that it was just a typoe.

peteryates: Done in #462

PR #462, opened 2023-11-16
Add spec ensuring viewBox is properly capitalised

peteryates:
XML (and therefore SVG) is case sensitive and viewBox is the correct attribute name, based on @aliuk2012's suggestion.

Refs #459, #461


lfdebrux (approved):
Tested locally, works for me!

PR #464, opened 2023-11-16
Update default navigation and menu button labels

### LINKED ISSUE #460: Default content for header labels does not match latest version of GOV.UK Frontend
### Description of issue

Another small regression against GOV.UK Frontend; I noticed that the default ARIA labels for the header buttons don't match those in GOV.UK Frontend v4 and up:

https://github.com/x-govuk/govuk-components/blob/30340f72ecb42086fc8ae641c3abef1a751f4414/lib/govuk/components/engine.rb#L90-L91

See 'Navigation menu' and 'Show or hide navigation menu' versus

https://github.com/alphagov/govuk-frontend/blob/v4.7.0/package/govuk/components/header/template.njk#L65-L66

'Menu' and 'Show or hide menu'.

It turns out the content was changed in the v4 release, as the inclusion of 'Navigation' in the label made some screen readers announcements sound a bit redundant. From the original commit:

> Without this change, assistive technologies report this as 'Navigation menu navigation'.
https://github.com/alphagov/govuk-frontend/commit/aed4b3e60f991d47595f4af7a7d03c8ff7f22455

### Workaround

Consumers of the `govuk_components` gem can easily change the default content across their app, so that's good! :D

```ruby
Govuk::Components.configure do |config|
  config.default_header_navigation_label = "Menu"
  config.default_header_menu_button_label = "Show or hide menu"
end
```

### Suggested fix

It would be nice if the defaults in the gem could be updated, so users get the benefit of the improved ARIA labels on all services that use this gem.

peteryates:
These had fallen behind govuk-frontend, this change brings them back in line. Thank you for spotting @lfdebrux!

Fixes #460


lfdebrux (approved):
Many thanks!

PR #465, opened 2023-11-26
Add visually hidden prefix and suffix text to links and buttons

peteryates:
It would be nice to be able to simply add visually hidden text to links and buttons without having to manually construct the HTML. It comes in really handy when working in dense admin interfaces where there might just be enough space for 'View', but we want to give screenreader users and idea of _where_ the link will take them.

## Work so far

- Split out `#extract_button_link_args`
- Rework and simplify the link and button helpers
- Add `govuk_visually_hidden` helper which makes creating visually hidden text easy
- Add keyword arguments `visually_hidden_prefix` and `visually_hidden_suffix` which injects the visually hidden text before or after the link or button text

## Example use

With keyword arguments:

```erb
<%= govuk_link_to("Delete", "#", visually_hidden_suffix: "first item")  %>
```

With a block:

```erb
<%= govuk_link_to("#") do %>
  Delete <%= govuk_visually_hidden("first item") %>
<% end %>
```

Both will output:

```html
<a href="#" class="govuk-link">Delete<span class="govuk-visually-hidden"> first item</span></a>
```

Prefixes work in the same way:

```erb
<%= govuk_link_to("Mr Jones", "#", visually_hidden_prefix: "Remove")  %>
```

```html
<a href="#" class="govuk-link"><span class="govuk-visually-hidden">Remove </span>Mr Jones</a>
```

### DIFF
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -34,7 +34,7 @@ module GovukLinkHelper
   end
 
   def govuk_button_link_to(name, href = nil, new_tab: false, disabled: false, inverse: false, secondary: false, warning: false, **kwargs, &block)
-    button_args = extract_button_args(new_tab: new_tab, disabled: disabled, inverse: inverse, secondary: secondary, warning: warning, **kwargs)
+    button_args = extract_button_link_args(new_tab: new_tab, disabled: disabled, inverse: inverse, secondary: secondary, warning: warning, **kwargs)
 
     if block_given?
       link_to(block.call, href, **button_args)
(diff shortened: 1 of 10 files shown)


peteryates:
@frankieroberto sorry yeah, didn't quite get round to actually making the changes yet, just some minor refactoring to pave the way :)

> ```
> govuk_link_to("Delete", "#", visually_hidden_suffix: "first item")
> ```

☝🏽 This first example is exactly what I had in mind, but I think there's definitely scope for a `govuk_visually_hidden` helper too.

frankieroberto:
Here's how it might look using a nested helper... 😬  Unfortunatately due to needing to concatenate a string with the visually-hidden tag, I think you'd also need to use `safe_join`?

```ruby
govuk_link_to(safe_join(["Delete, govuk_visually_hidden(' first item')]), "#")
```

or you could concatenate anyway and mark the output as safe (so long as the string isn't user-submitted), but this feels even worse:
```ruby
govuk_link_to("Delete#{govuk_visually_hidden(' first item')}".html_safe, "#")
```

...in either case, probably the block syntax is better?

app/helpers/govuk_link_helper.rb
```diff
@@ -122,6 +122,19 @@ def extract_button_classes(inverse: false, secondary: false, warning: false)
...
+  def build_text(original, visually_hidden_prefix:, visually_hidden_suffix:, &block)
+    text = (block_given?) ? block.call : original
+
+    safe_join(
+      [
+        govuk_visually_hidden(visually_hidden_prefix),
+        text,
+        govuk_visually_hidden(visually_hidden_suffix),
+      ].compact,
+      " "
```
peteryates, line 135: This currently adds space between the visually hidden span and the text - I'll switch it to putting the space inside the visually hidden span, I think that's the right approach.

peteryates:
@frankieroberto yeah the reason i want to add these as keywords is because remembering where to add the space gets a bit clunky. Currently it's wrong, I'll amend it so it's _inside_ the visually hidden span next (it's currently outside).

app/helpers/govuk_link_helper.rb
```diff
@@ -124,16 +124,11 @@ def brand
   end
 
   def build_text(original, visually_hidden_prefix:, visually_hidden_suffix:, &block)
+    prefix = (visually_hidden_prefix.present?) ? visually_hidden_prefix + " " : nil
     text = (block_given?) ? block.call : original
+    suffix = (visually_hidden_suffix.present?) ? " " + visually_hidden_suffix : nil
```
peteryates, line 129: I think this can be made a bit more elegant.

PR #466, opened 2023-11-27
Update to use new combined logo svg

frankieroberto:
The `govuk-frontend` v5.0.0-beta.2 [contains some changes to the way the logo is rendered](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0-beta.2).  The SVG image now contains both the crown and the GOV.UK text.

This means that for the `govuk-component` gem, we can no longer support custom text alongside the crown. Users will either have to use the default crown + GOV.UK text, or replace both.  Luckily, it doesn’t seem that many people used this feature.

Consequences for the ruby implementation are that:

* The `logotype` property of the header component has been dropped
* The `default_header_logotype` config option has been dropped (previously this defaulted to `GOV.UK`)

I've also removed the `crown` boolean property of the header component, as this is no longer needed.

Users who need to specify an alternative logo can continue to use the `custom_logo` slot, but this will now need to contain both any logo and any text.

This PR also removes the X-UA-Compatible meta tag, as recommended by the beta.2 release notes.


paulrobertlloyd:
Mixed feelings about this, but I guess this is **GOV.UK** frontend, and if you want to change an organisation name and logo, you’ll likely want to make other changes to the header as well. And besides, you can still add your own HTML to the `custom_logo` slot.

frankieroberto:
> Mixed feelings about this, but I guess this is **GOV.UK** frontend, and if you want to change an organisation name and logo, you’ll likely want to make other changes to the header as well. And besides, you can still add your own HTML to the `custom_logo` slot.

@paulrobertlloyd Yes, we pretty much have to do this seeing as the GOV.UK logo text is now embedded in the SVG.

That said, I don't think we know of anyone using the crown with alternative text?

And if you're using an alternative logo + text you might well need some custom css too to adjust the vertical offset. (the govuk one uses `-3px` to align the logo to the product name).

peteryates (approved):
Looks great.

colinrotherham:
Brill to see this 🙌 

We added a tiny fix via `aria-label` as the SVG `<title>`  wasn't announced in VoiceOver/Safari

* https://github.com/alphagov/govuk-frontend/pull/4497

If you can please add the fix to your template too, it all helps

frankieroberto:
@colinrotherham thanks - fixed in #467

PR #476, opened 2023-11-29
Fix spelling of compatible

peteryates:


frankieroberto (approved):
Nice. We'll have to note this as another breaking change, as I think this was introduced in a previous version, I think?

Not sure how widespread its use is though?!

peteryates:
This is yet unreleased, I merged the original branch to `main` first and then merged main into this branch.

PR #481, opened 2023-12-10
Append card title to summary list actions

### LINKED ISSUE #479: Update visually hidden text in action links on summary cards
We should apply the changes from https://github.com/alphagov/govuk-frontend/pull/3961

This means that, where a summary card is used, the actions in both the title or the summary list should have the card title added to the visually-hidden text within brackets.

This means that instead of

```html
<a href="" class="govuklink">Change<span class="govuk-visually-hidden"> address</span></a>
```

you have

```html
<a href="" class="govuklink">Change<span class="govuk-visually-hidden"> address (Person 1)</span></a>
```

peteryates:
This change appends the card title to action links when rendered within a summary card. It follows [the upstream change](https://github.com/alphagov/govuk-frontend/pull/3961) introduced in govuk-frontend version 5.

In the simplest case it works like the reference implementation:

```erb
<%= govuk_summary_list(card: { title: "Shiny new widget" }) do |sl| %>
  <%= sl.with_row do |r| %>
    <%= r.with_key(text: "Price") %>
    <%= r.with_value(text: "£19.99") %>
    <%= r.with_action(href: "/widgets/1/edit", text: "Modify", visually_hidden_text: "this widget's price") %>
  <% end %>
<% end %>
```

Here, the rendered HTML will now include the card title in the visually hidden text on the action:

```html
<dl class="govuk-summary-list">
  <div class="govuk-summary-list__row">
    <dt class="govuk-summary-list__key">Price</dt>
    <dd class="govuk-summary-list__value">£19.99</dd>
    <dd class="govuk-summary-list__actions">
      <a class="govuk-link" href="/a">Modify<span class="govuk-visually-hidden"> this widget's price (Shiny new widget)</span></a>
    </dd>
  </div>
</dl>
```

Additionally the suffix text can be set manually. This allows it to be set without passing in `card: { ... }` when initialising the summary list. It might be useful when rendering the summary list within an outer card. It is set using the `action_suffix` keyword.

```erb
<%= govuk_summary_card(title: "Shiny new widget") do %>
  <%= govuk_summary_list(action_suffix: "Shiny new widget") do |sl| %>
    <%= sl.with_row do |r| %>
      <%= r.with_key(text: "Price") %>
      <%= r.with_value(text: "£19.99") %>
      <%= r.with_action(href: "/widgets/1/edit", text: "Modify", visually_hidden_text: "this widget's price") %>
    <% end %>
  <% end %>
<% end %>
```

This will output the same HTML as the above example.

When both the `card` and `action_suffix` arguments are provided the `action_suffix` takes precedence, which offers a little more flexibility as there might be some cases other text would make more sense than the card's title.

Fixes #479, #480

### DIFF
--- a/app/components/govuk_component/summary_list_component.rb
+++ b/app/components/govuk_component/summary_list_component.rb
@@ -1,22 +1,22 @@
 module GovukComponent
   class SummaryListComponent < GovukComponent::Base
-    attr_reader :borders, :actions, :card, :action_suffix
+    attr_reader :borders, :actions, :card, :visually_hidden_action_suffix
 
     renders_many :rows, ->(classes: [], html_attributes: {}, &block) do
       GovukComponent::SummaryListComponent::RowComponent.new(
         show_actions_column: @show_actions_column,
-        action_suffix: action_suffix || card&.title,
+        visually_hidden_action_suffix: visually_hidden_action_suffix || card&.title,
         classes: classes,
         html_attributes: html_attributes,
         &block
       )
     end
 
-    def initialize(rows: nil, actions: true, borders: config.default_summary_list_borders, card: {}, action_suffix: nil, classes: [], html_attributes: {})
-      @borders             = borders
-      @show_actions_column = actions
-      @card                = GovukComponent::SummaryListComponent::CardComponent.new(**card) if card.present?
-      @action_suffix       = action_suffix
+    def initialize(rows: nil, actions: true, borders: config.default_summary_list_borders, card: {}, visually_hidden_action_suffix: nil, classes: [], html_attributes: {})
+      @borders                       = borders
+      @show_actions_column           = actions
+      @card                          = GovukComponent::SummaryListComponent::CardComponent.new(**card) if card.present?
+      @visually_hidden_action_suffix = visually_hidden_action_suffix
 
       super(classes: classes, html_attributes: html_attributes)
 
--- a/app/components/govuk_component/summary_list_component.rb
+++ b/app/components/govuk_component/summary_list_component.rb
@@ -1,21 +1,22 @@
 module GovukComponent
   class SummaryListComponent < GovukComponent::Base
-    attr_reader :borders, :actions, :card
+    attr_reader :borders, :actions, :card, :action_suffix
 
     renders_many :rows, ->(classes: [], html_attributes: {}, &block) do
       GovukComponent::SummaryListComponent::RowComponent.new(
         show_actions_column: @show_actions_column,
-        card_title: card&.title,
+        action_suffix: action_suffix || card&.title,
         classes: classes,
         html_attributes: html_attributes,
         &block
       )
     end
 
-    def initialize(rows: nil, actions: true, borders: config.default_summary_list_borders, card: nil, classes: [], html_attributes: {})
+    def initialize(rows: nil, actions: true, borders: config.default_summary_list_borders, card: {}, action_suffix: nil, classes: [], html_attributes: {})
       @borders             = borders
       @show_actions_column = actions
-      @card                = GovukComponent::SummaryListComponent::CardComponent.new(**card) unless card.blank?
+      @card                = GovukComponent::SummaryListComponent::CardComponent.new(**card) if card.present?
+      @action_suffix       = action_suffix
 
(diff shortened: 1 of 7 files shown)


app/components/govuk_component/summary_list_component/action_component.rb
```diff
@@ -10,8 +10,9 @@ def initialize(href: nil, text: 'Change', visually_hidden_text: false, classes:
 
     super(classes: classes, html_attributes: html_attributes)
 
-    @href = href
-    @text = text
+    @href       = href
+    @text       = text
+    @action_suffix = action_suffix
```
peteryates, line 15: Will fix the wonky alignment 🤦🏽‍♂️

app/components/govuk_component/summary_list_component.rb
```diff
@@ -1,20 +1,22 @@
 module GovukComponent
   class SummaryListComponent < GovukComponent::Base
-    attr_reader :borders, :actions, :card
+    attr_reader :borders, :actions, :card, :action_suffix
 
     renders_many :rows, ->(classes: [], html_attributes: {}, &block) do
       GovukComponent::SummaryListComponent::RowComponent.new(
         show_actions_column: @show_actions_column,
+        action_suffix: action_suffix || card&.title,
```
frankieroberto, line 8: I think I would be tempted to call the attribute `card_title` rather than `action_suffix`, just to be a bit clearer what you’d set it to? `action_suffix` might be a bit ambiguous, as it's not clear it's adding visually-hidden text, plus there’s already visually-hidden suffixes within the action links? 🤔 

```suggestion
    attr_reader :borders, :actions, :card, :card_title

    renders_many :rows, ->(classes: [], html_attributes: {}, &block) do
      GovukComponent::SummaryListComponent::RowComponent.new(
        show_actions_column: @show_actions_column,
        card_title: card_title || card&.title,
```

app/components/govuk_component/summary_list_component/action_component.rb
```diff
@@ -1,7 +1,7 @@
 class GovukComponent::SummaryListComponent::ActionComponent < GovukComponent::Base
-  attr_reader :href, :text, :visually_hidden_text, :attributes, :classes
+  attr_reader :href, :text, :visually_hidden_text, :action_suffix, :attributes, :classes
```
frankieroberto, line 2: Think I'd call this `card_title` too, to better match [the nunjucks implementation](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/summary-list/template.njk#L7)?

```suggestion
  attr_reader :href, :text, :visually_hidden_text, :card_title, :attributes, :classes
```

frankieroberto (commented):
Looks good - thanks for getting this working, I tried but couldn't easily figure out how to pass the card reference from the summary card to the summary list component.

I’m not sure about `action_suffix` as a term, but if you can convince me there’s other use-cases for this other than card titled, then it's probably fine. 

The way that the card title is hardcoded to appear within `()` parentheses make it hard to predict what else you might use it for, especially as you can already add visually-hidden text?

The one other thing I don't think this PR does yet is to also add the card title as visually-hidden text to any actions within the header of the card?

I think this is tricker though as currently those actions are just slots that you render a `govuk_link_to` in, rather than being built from parameters?

app/components/govuk_component/summary_list_component.rb
peteryates: I had it as `card_title` originally but it felt weird when rendering it standalone, so changed it. It feels weird to me that an argument called `card_title` doesn't set the title of a card.

frankieroberto: Hmm good point. I guess conceptually you're telling the summary list what the card_title _is_, so that it can be used as visually-hidden text. But maybe that's confusing...

Today in "naming things is hard"... 😂

app/components/govuk_component/summary_list_component/action_component.rb
peteryates: The alignment was because I did a find and replace for `card_title` to `action_suffix` 😅

peteryates:
Merging this. Went with `visually_hidden_action_suffix` for now, but it's undocumented and only intended for internal use, so we can change if needed.

PR #483, opened 2023-12-13
Add warning when action/controller style links used

peteryates:
Support for this format was dropped in version 5 and upgrading/reverting is covered in the linked release notes.


frankieroberto (approved):
Nice.

PR #484, opened 2023-12-18
Fix the breadcrumb helper and add extra tests

peteryates:
This bug was kindly reported by @paulrobertlloyd.

| Before | After |
| ------ | ------ |
| Screenshot from 2023-12-18 16-01-56 | Screenshot from 2023-12-18 16-02-15 |


frankieroberto (approved):
Nice spot!

PR #486, opened 2023-12-24
Documentation design update

paulrobertlloyd:
This PR does a few things, increasingly more ambitious/contentious!

- Fixes the logo in the header now that we need to support the new v5 way of doing things
- Removes the inverse styles that are no longer needed as supplied by v5
- Uses the Masthead component styles from the GOVUK Prototype Components project
- Adds a modifier class to the header on the index page to account for the masthead
- Increases the overall width of the page container, and adds sub-navigation to all pages

Happy to break apart into separate PRs, but hopefully nothing in here is too contentious.

And if you’re happy with these changes, I’ll go off and do the same for the Form Builder docs. 🎁

Screenshot of Getting started page showing updated logo in the header and sub navigation with the current page selected.


peteryates (approved):
This is great, everything's super clear and the guide is _so much easier_ to navigate now.

PR #488, opened 2023-12-26
Add `cannot_start_yet` argument to the task list status component

peteryates:
It provides a nicer way to add the class `govuk-task-list__status--cannot-start-yet` than manually providing it.

### DIFF
--- a/app/components/govuk_component/task_list_component/item_component.rb
+++ b/app/components/govuk_component/task_list_component/item_component.rb
@@ -40,6 +40,10 @@ module GovukComponent
     end
 
     def call
+      if href.presence && status_content.cannot_start_yet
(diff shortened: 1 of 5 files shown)


frankieroberto (approved):
Nice! Wonder if we should throw an error if this is set to true and an href is present?

peteryates:
That sounds sensible to me. I don't think there's ever a reason for a row to be 'cannot start' and have a href 🤔 

I guess a warning message in the log could work too if we want to give _some_ degree of flexibility.

frankieroberto:
I think I would go with raising an error as warnings in the logs can be missed or ignored - the error could also suggest that if you _really_ wanted to do it, you could add the class manually?

(We could also drop the error in future if someone explains a good reason to have a link for something that can’t be started yet?)

PR #490, opened 2023-12-27
Documentation design update (supplemental)

paulrobertlloyd:
Following on from #486, a few supplemental tweaks:

- Consolidate header styles (should have placed all the header styles used for the guide in the same file; had I done initially I would have seen a rule I should have removed)
- Consistent syntax and the navigation Slim templates, to hopefully be a bit clearer
- Use the related navigation component styles from the GOVUK Prototype Components project for related information


peteryates (approved):
Looks great, thanks.

PR #497, opened 2024-01-28
New tabs with descriptive text

### LINKED ISSUE #489: Improvements to `new_tab` behaviour
The `new_tab` argument on links adds the `target` and `rel` attributes to links so they open in a new tab, but it doesn't add the recommended text suffix `(opens in new tab)` to the link.

I think we should add the following behaviour:

* automatically append the text `(opens in new tab)` to links that have `new_tab: true`
* allow the text to be configured using the `default_link_new_tab_text` config option
* allow the behaviour to be suppressed by setting `default_link_new_tab_text` to `nil` or `false`
* add a new argument to link helpers called `visually_hidden_new_tab_text` or similar, which, if true, will append `<span class="govuk-visually-hidden">(opens in new tab)</span>` instead

peteryates:
The GOV.UK Design System says:

> If you need a link to open in a new tab - for example, to stop the user losing information they’ve entered into a form - then include the words ‘opens in new tab’

However, in the examples the new tab text is surrounded by parentheses.

To reduce repetition here we are changing the behaviour of links that open in a new tab so that the text is automatically inserted.

The text defaults to `(link opens in new tab)` but this can be overridden using the `default_link_new_tab_text` configuration option.

The behaviour can be suppressed entirely by setting this value to `nil`.

On a link-by-link basis, the new tab text can be overridden by passing a string instead of true to `new_tab`, so calling:

```erb
<%= govuk_link_to("hello", "/some/page", new_tab: "(opens in new window)") %>
```

will generate a link that has the overridden text as a suffix.

If we want to visually hide this text it could be called with:

```erb
<%= govuk_link_to("hello", "/some/page", new_tab: govuk_visually_hidden("(opens in new window)")) %>
```

## Changes made

- Remove new_tab keyword from `govuk_mail_to`
- Add new tab text to links that open in new tab
- Add a guide entry covering the new tab behaviour
- Improve the argument assignment in links

Fixes #489


spec/helpers/govuk_link_helper_spec.rb
```diff
@@ -80,11 +80,40 @@ def url_for(path)
...
+
+      context "when new_tab: '(opens in a new window)'" do
+        let(:kwargs) { { new_tab: '(opens in a new window)' } }
+
+        specify "the link has the new tab attributes and the provided text is appended" do
+          expect(subject).to have_tag("a", text: "hello (opens in a new window)", with: expected_new_tab_attributes)
+        end
+      end
+
+      context "when new_tab: nil" do
```
frankieroberto, line 102: I think I would expect `nil` to do the same as `false` - ie don't open in a new tab, as nil is usually falsy. 

For opening in a new tab but with no text appended, an empty string `""` feels more intuitive to me?

spec/helpers/govuk_link_helper_spec.rb
```diff
@@ -345,20 +366,40 @@ def url_for(path)
       end
     end
 
-    context "when new_tab: true" do
-      let(:kwargs) { { new_tab: true } }
+    describe "new_tab" do
```
frankieroberto, line 369: Is this a duplicate of the above, or is this for a different helper? (unclear from the diff)

spec/helpers/govuk_link_helper_spec.rb
```diff
@@ -80,11 +80,40 @@ def url_for(path)
...
+      let(:expected_new_tab_attributes) { { href: "/world", class: "govuk-link", target: "_blank", rel: "noreferrer noopener" } }
 
-      specify "the new tab attributes are present on the link" do
-        expect(subject).to have_tag("a", text: "hello", with: { href: "/world", class: "govuk-link", target: "_blank", rel: "noreferrer noopener" })
+      context "when new_tab: true" do
+        let(:kwargs) { { new_tab: true } }
+
+        specify "the link has the new tab attributes and the default 'new tab' text is appended" do
+          expect(subject).to have_tag("a", text: "hello (opens in new tab)", with: expected_new_tab_attributes)
```
frankieroberto, line 90: Up to you, but I always think these tests are easier to follow if the expectations are explicitly defined inline, at the expense of duplication?  Could also only test the attributes that are different from the other link tests which are tested further up?

```suggestion
          expect(subject).to have_tag("a", text: "hello (opens in new tab)", with: { target: "_blank", rel: "noreferrer noopener" })
```

spec/helpers/govuk_link_helper_spec.rb
```diff
@@ -80,11 +80,40 @@ def url_for(path)
...
     end
 
-    context "when new_tab: true" do
-      let(:kwargs) { { new_tab: true } }
+    describe "new_tab" do
+      let(:expected_new_tab_attributes) { { href: "/world", class: "govuk-link", target: "_blank", rel: "noreferrer noopener" } }
 
-      specify "the new tab attributes are present on the link" do
-        expect(subject).to have_tag("a", text: "hello", with: { href: "/world", class: "govuk-link", target: "_blank", rel: "noreferrer noopener" })
+      context "when new_tab: true" do
```
frankieroberto, line 86: Is there a test for `new_tab: false` too?

guide/content/helpers/link.slim
```diff
@@ -45,6 +45,26 @@ markdown:
...
+  caption: "Links that open in a new tab",
+  code: govuk_link_to_new_tab) do
+
+  markdown:
+    You can use the `new_tab` parameter to automatically set the `target` and
+    `rel` [attributes as suggested by the design system][0].
+
+    When the link text is passed in as a string argument, 'opens in new tab' is
+    added to the end automatically. No suffix is added when the helper is used
+    in block mode.
```
frankieroberto, line 58: This confused me at first, as I mis-read as describing when the new tab text is added as a string.

Is it worth having a few different examples?

1. `new_tab: true` for the default text
2. `new_tab: "(opens in a new window)" to override the default
3. `new_tab: govuk_visually_hidden("(opens in a new tab)")` to visually-hide it (example could have a preceding line saying "The following links open in a new tab" to follow the Design System guidance?

Could also mention how to turn off the text altogether, but maybe not include an example, as this pattern isn't recommended?!

spec/helpers/govuk_link_helper_spec.rb
peteryates: Yeah, this is for `govuk_button_link_to`. At some point I'll reorganise these into shared contexts.

frankieroberto: Ah cool. Wasn’t obvious from the diff. Moving to separate files might be enough. I’m never convinced by shared contexts as it’s hard to work out where they’re defined.

spec/helpers/govuk_link_helper_spec.rb
frankieroberto: @peteryates any thoughts on this? I know nil vs blank is a bone of contention across many languages… 😂

peteryates: Hmm not sure I agree _entirely_ - I've changed it so:

* `new_tab: true` creates link that opens in a new tab with the default text suffix (set in config)
* `new_tab: "(some text)"` creates link that opens in new tab with ` (some text)` on the end
* `new_tab: nil` creates a link that opens in a tab without any extra text
* `new_tab: ""` creates a link that opens in a tab without any extra text

spec/helpers/govuk_link_helper_spec.rb
peteryates: There is now!

guide/content/helpers/link.slim
peteryates: Agreed and implemented.

spec/helpers/govuk_link_helper_spec.rb
peteryates: I'll leave it for now just because the attributes are a bit wordy and the variable is defined within the parent group.

spec/helpers/govuk_link_helper_spec.rb
frankieroberto: Ok good compromise! 😃

spec/helpers/govuk_link_helper_spec.rb
peteryates: Yeah that's the issue I have with them too, I want something I can jump to in my editor but the string names don't help.

frankieroberto (approved):

PR #503, opened 2024-02-05
GOV.UK Frontend v5.1.0

peteryates:
- Upgrade to govuk-frontend version 5.1.0
- Remove aria-labelledby from accordion sections
- Switch the crown 👑
- Replace the slightly-pointless accordion template


asmega:
@peteryates would it not be easier to link to the image rather than embed it? it would save having to update it on each occurrence but a slightly bigger impact would be the browser can cache it after the initial extra http call

frankieroberto:
@asmega we’ve followed the markup from GOVUK Frontend which has the SVG inline. I think this is for accessibility as inline SVGs are easier to style for things like high contrast mode.

PR #506, opened 2024-02-09
Updates to Tabs documentation

frankieroberto:
When implementing this I initially didn't spot that the the tab slots would accept a block. Feel like this is probably the default use case (as tabs normally contain a fair amount of content) so worth highlighting this a bit more.

Also added some documentation on the `title` argument, and renamed the local variable from `c` to `tabs`.

### DIFF
--- a/guide/lib/examples/tabs_helpers.rb
+++ b/guide/lib/examples/tabs_helpers.rb
@@ -3,11 +3,11 @@ module Examples
     def tabs_normal
       <<~TABS
         = govuk_tabs(title: "Contents") do |tabs|
-          - tabs.with_tab(label: "Monday")
-            | Monday’s child is fair of face
-          - tabs.with_tab(label: "Tuesday")
-            | Tuesday’s child is full of grace
-          - tabs.with_tab(label: "Wednesday", text: "Wednesday’s child is full of woe")
+          - tabs.with_tab(label: "Text", text: "This was set using a text argument")
+          - tabs.with_tab(label: "Inline block") { "This was set using an inline block of content" }
+          - tabs.with_tab(label: "Regular block") do
+            p This was set using a block of HTML.
+            p Use this style if you need complex markup within your tab.
       TABS
     end
   end
--- a/guide/lib/examples/tabs_helpers.rb
+++ b/guide/lib/examples/tabs_helpers.rb
@@ -2,11 +2,12 @@ module Examples
   module TabsHelpers
     def tabs_normal
       <<~TABS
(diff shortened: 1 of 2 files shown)


guide/lib/examples/tabs_helpers.rb
```diff
@@ -2,11 +2,12 @@ module Examples
...
-          - c.with_tab(label: "Monday", text: "Monday’s child is fair of face")
-          - c.with_tab(label: "Tuesday", text: "Tuesday’s child is full of grace")
-          - c.with_tab(label: "Wednesday")
-            | Wednesday’s child is full of woe
+        = govuk_tabs(title: "Contents") do |tabs|
+          - tabs.with_tab(label: "Monday")
+            | Monday’s child is fair of face
+          - tabs.with_tab(label: "Tuesday")
+            | Tuesday’s child is full of grace
+          - tabs.with_tab(label: "Wednesday", text: "Wednesday’s child is full of woe")
```
peteryates, line 10: I just had a little play and remember why I did it like this now. There's a weird bit of behaviour in the guide that doesn't manifest in Rails where things added in blocks aren't cleared for the next block, resulting in content doubling up like this:

image

Maybe changing the example to something like this would make things clearer?

```suggestion
        = govuk_tabs(title: "Contents") do |tabs|
          - tabs.with_tab(label: "Text", text: "This was set using a text argument")
          - tabs.with_tab(label: "Inline block") { "This was set using an inline block of content" }
          - tabs.with_tab(label: "Regular block") do
            p This was set using a block of HTML.
            p Use this style if you need complex markup within your tab.
```

frankieroberto: Weird! Any idea why this happens?

peteryates: No, I tried to solve it with no luck. I'm sure it's because in the guide we're basically setting up a _just enough Rails to perform_ pile of components that aren't really geared towards working outside of Rails. It pulls in bits of ActionView and sets up a `Rails::Application`, and combines that with stuff from Nanoc and then adds in the components and helpers from the library.

To top off the weirdness, all the examples are strings in methods, which are piped through Slim to create the output. I like this approach because the guide remains static _and_ it uses the library itself, so it can't be out of date. But, I'm the first to admit it's hardly the most robust thing ever 😅

peteryates: @frankieroberto happy with a change similar to the suggested one? If so I'll start getting a 5.1.0 beta release together.

peteryates (approved):

PR #511, opened 2024-02-23
Move custom meta content above licence in footer

peteryates:
It appears that the custom meta content should be above the licence rather than underneath it.

The [examples in the design system samples show it](https://govuk-frontend-review.herokuapp.com/components/footer) but it's not really covered by [the official documentation](https://design-system.service.gov.uk/components/footer/) (which was what the library was based on in the early days).

| Before | After |
| ------ | ------- |
| Screenshot from 2024-02-23 11-53-34 | Screenshot from 2024-02-23 11-56-58 |

### DIFF
--- a/guide/lib/examples/footer_helpers.rb
+++ b/guide/lib/examples/footer_helpers.rb
@@ -35,7 +35,8 @@ module Examples
       <<~FOOTER_META_HTML
         = render GovukComponent::FooterComponent.new do |footer|
           - footer.with_meta_html do
-            | Built with love by x-govuk.
+            | Built with love by
+            = govuk_footer_link_to("X-GOVUK", "https://x-govuk.github.io/")
       FOOTER_META_HTML
     end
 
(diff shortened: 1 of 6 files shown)


guide/lib/examples/footer_helpers.rb
```diff
@@ -34,10 +34,8 @@ def footer_meta_complex_items
     def footer_with_custom_meta_html
       <<~FOOTER_META_HTML
         = render GovukComponent::FooterComponent.new do |footer|
-          - footer.meta_html do
-            .govuk-footer__meta-custom class="govuk-\!-margin-top-1"
-              | Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam
-                a porta purus. Fusce faucibus aliquam massa sed eleifend.
+          - footer.with_meta_html do
+            | Built with love by x-govuk.
```
paulrobertlloyd, line 38: Might be useful to show including a link in this example (which would need the `govuk-footer__link` applied); not sure what the correct way of doing that would be, off the top of my heading, but I guess using the `govuk_link` helper.

```suggestion
            | Built with love by X-GOVUK.
```

paulrobertlloyd:
Good spot! Think we can improve the example a little, but nothing more to add.

guide/lib/examples/footer_helpers.rb
peteryates: Good point. I think there's room for a `#govuk_footer_link_to` helper as the ones in the example don't have the `govuk-link` class.

PR #513, opened 2024-03-01
Add data-module and prevent-double-click to buttons

misaka:
These features were present in the previous version of the link helper but didn't make it to the rewrite.

### DIFF
--- a/spec/helpers/govuk_link_helper_spec.rb
+++ b/spec/helpers/govuk_link_helper_spec.rb
@@ -327,7 +327,7 @@ RSpec.describe(GovukLinkHelper, type: 'helper') do
     subject { govuk_button_link_to("hello", "/world", **kwargs) }
 
     specify "renders a link styled as a button with the correct class" do
-      expect(subject).to have_tag("a", text: "hello", with: { href: "/world", class: "govuk-button", "data-module": "govuk-button-to" })
+      expect(subject).to have_tag("a", text: "hello", with: { href: "/world", class: "govuk-button", "data-module": "govuk-button" })
     end
 
     context "when a custom brand is set" do
@@ -344,7 +344,7 @@ RSpec.describe(GovukLinkHelper, type: 'helper') do
       end
 
       specify "attributes are branded" do
-        expect(subject).to have_tag("a", text: "hello", with: { href: "/world", class: "globex-corp-button", "data-module": "globex-corp-button-to" })
+        expect(subject).to have_tag("a", text: "hello", with: { href: "/world", class: "globex-corp-button", "data-module": "globex-corp-button" })
       end
     end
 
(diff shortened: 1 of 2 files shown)


frankieroberto:
@misaka is it useful to prevent double click on links styled as buttons? These shouldn’t be affected by the issue than can occur on real buttons (ie a form being submitted twice).

That said, maybe it doesn’t do any harm?

misaka:
> @misaka is it useful to prevent double click on links styled as buttons? These shouldn’t be affected by the issue than can occur on real buttons (ie a form being submitted twice).
> 
> That said, maybe it doesn’t do any harm?

Ah, I thought I was just copying what was there previously, but you're right ... I got a bit overzealous. And like you say, it's not really useful for links.

I'll pull it back out, thanks @frankieroberto!

spec/helpers/govuk_link_helper_spec.rb
```diff
@@ -327,7 +327,25 @@ def url_for(path)
...
+
+      around do |ex|
+        Govuk::Components.configure do |conf|
+          conf.brand = custom_brand
+        end
+
+        ex.run
+
+        Govuk::Components.reset!
+      end
```
peteryates, line 344: Way nicer than `before` and `after` - I will start using this too!

peteryates:
Thanks for this @misaka, looking good.

I had a check and it looks @frankie's PR #449 to add it came after the new links/button branch had been opened, and I missed it from any rebasing 🙈

misaka:
Updated without the double-click protection on button links, and with a small (but critical!) fix to the data-module generated. I've tested this branch with our service.

peteryates (approved):
Looks great, thanks @misaka 😄

misaka:
Actually, I did just find somewhere where preventing a double-click on a link could marginally be useful.

We have a download link that generates a CSV and sends it to the user. Not as severe as protecting against double-POSTs but if generating the file is heavy or if we simply want to protect the user against downloading too much data / being confused by the duplicate filenames, this could be useful here.

frankieroberto:
It looks like the govuk-button component will add a `data-module="govuk-button"` to links styled as button, but does not support prevendDoubleClick on links styled as buttons?

See https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/button/template.njk

misaka:
> It looks like the govuk-button component will add a `data-module="govuk-button"` to links styled as button, but does not support prevendDoubleClick on links styled as buttons?
> 
> See [alphagov/govuk-frontend@`main`/packages/govuk-frontend/src/govuk/components/button/template.njk](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/button/template.njk?rgh-link-date=2024-03-01T14%3A23%3A11Z)

I see what you mean looking at the code. However, if I set `data-prevent-double-click="true"` manually on the `a`, it does work. I've added a screenshot of the element with the event on it, which I've confirmed works.

I'm not familiar enough with `govuk-frontend` to say whether this is a Good Idea<sup>tm</sup> or a Bad Idea<sup>tm</sup>, happy to add it if you think it's good, or drop it if not.

image

peteryates:
Will merge this now, I've raised #517 to add the data-module to investigate adding `data-module` to links styled as buttons too.

PR #518, opened 2024-03-02
Build copyright when rendering instead of at init

### LINKED ISSUE #512: Previews for Components
The [previews documentation](https://viewcomponent.org/guide/previews.html) details how to setup previews similar to action mailer previews, in a contained environment.

The `GovukComponent::FooterComponent` component breaks the guidelines of not relying on the `controller` during initialisation.

CleanShot 2024-02-28 at 12 54 01

Upon investigation, the [govuk_footer_link_to](https://github.com/x-govuk/govuk-components/blob/main/app/components/govuk_component/footer_component.rb#L70) calls within [build_meta_links](https://github.com/x-govuk/govuk-components/blob/main/app/components/govuk_component/footer_component.rb#L25) are the source of this error.

The [link_to](https://github.com/x-govuk/govuk-components/blob/main/app/components/govuk_component/footer_component.rb#L85) call on line 85 seems to be fine though. My initial assumption is that `ViewComponent` defines `link_to` somewhere which is separate than the `link_to` referenced within our view helpers.

We have created a custom `FooterComponent`, in our repo [here](https://github.com/DFE-Digital/itt-mentor-services/pull/315), that inherits from `GovukComponent::FooterComponent`.

Could we move the presentation logic into the `erb` template itself?

peteryates:
ViewComponent raises a `ControllerCalledBeforeRenderError` when rendering previews because we're calling `#link_to` at the point the component is initialized, which is problematic because the view context only exists once the ViewComponent is passed to the Rails render pipeline.

Setting `copyright_text` and `copyright_url` during initialization but combining them into a link with `#link_to` when rendering the component should alleviate this.

Fixes #512

### DIFF
--- a/app/components/govuk_component/footer_component.rb
+++ b/app/components/govuk_component/footer_component.rb
@@ -5,7 +5,7 @@ class GovukComponent::FooterComponent < GovukComponent::Base
   renders_one :meta
   renders_one :navigation
 
-  attr_reader :meta_items, :meta_text, :meta_items_title, :meta_licence, :copyright, :custom_container_classes
+  attr_reader :meta_items, :meta_text, :meta_items_title, :meta_licence, :copyright_text, :copyright_url, :custom_container_classes
 
   def initialize(
     classes: [],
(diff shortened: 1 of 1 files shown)


frankieroberto:
@peteryates looks good to me! Anything stopping this from getting merged?

peteryates:
No, I think it's good to go.

frankieroberto (approved):

PR #519, opened 2024-03-02
Add before and after links slots to footer

peteryates:
These slots don't have any functionality but allow arbitrary html to be placed before/after the meta links.

* [x] Update the template so unnecessary whitespace isn't added when the slots aren't used
* [x] Make the testing a bit more robust


Nitemaeric (approved):
If it's possible to get this in, we'll adopt this over our current custom implementation.

peteryates:
@Nitemaeric sorry forgot about this, will get it merged and released today.

@frankieroberto this change look sensible to you?

frankieroberto (approved):
Looks fine to me, if there's a need for this.

Could rename to `content_before_meta_items` just to be consistent?

PR #520, opened 2024-03-05
Add list helper

### LINKED ISSUE #509: Lists
The GOV.UK Design System has [styles for lists](https://design-system.service.gov.uk/styles/lists/), but no Nunjucks macros for them.

However we could follow the convention we’ve established with [links](https://govuk-components.netlify.app/helpers/link/) and add a helper for them anyway?

Possible syntax:

#### Bullets

```erb
<%= govuk_list(type: "bullet") do |list| %>
  <%= list.with_item(text: "apples") %>
  <%= list.with_item(text: "oranges") %>
  <%= list.with_item(text: "pears") %>
<% end %>
```

#### Numbered

```erb
<%= govuk_list(type: "number") do |list| %>
  <%= list.with_item(text: "Delivery address.") %>
  <%= list.with_item(text: "Payment.") %>
  <%= list.with_item(text: "Confirmation.") %>
<% end %>
```

#### Bullets, spaced-out

```erb
<%= govuk_list(type: "bullets", spaced: true) do |list| %>
  <%= list.with_item do %>
    constructing, altering, repairing, extending, demolishing or dismantling buildings or structures (whether permanent or not), including offshore installation services
  <% end %>
  <%= list.with_item do %>
    constructing, altering, repairing, extending, demolishing of any works forming, or planned to form, part of the land, including (in particular) walls, roadworks, power lines, electronic communications equipment, aircraft runways, railways, inland waterways, docks and harbours
  <% end %>
<% end %>
```

You could also choose to use it like this:

```erb
<%= govuk_list(type: "bullet") do %>
  <li>apples</li>
  <li>oranges</li>
  <li>pears</li>
<% end %>
```

Is this helpful? Or overkill, for something you can do with plain HTML easily enough?

frankieroberto:
Add a helper for lists.

Resolves #509

### DIFF
--- a/app/helpers/govuk_list_helper.rb
+++ b/app/helpers/govuk_list_helper.rb
@@ -6,11 +6,13 @@ module GovukListHelper
 
     tag_type = (type == "number") ? "ol" : "ul"
 
-    html_classes = ["#{brand}-list"]
-    html_classes << "#{brand}-list--bullet" if type == "bullet"
-    html_classes << "#{brand}-list--number" if type == "number"
-    html_classes << "#{brand}-list--spaced" if spaced
-    html_classes += Array.wrap(classes)
+    html_classes = class_names(
+      "#{brand}-list",
+      *classes,
+      "#{brand}-list--bullet" => (type == "bullet"),
+      "#{brand}-list--number" => (type == "number"),
+      "#{brand}-list--spaced" => spaced
+    )
 
     if block_given?
       content_tag(tag_type, class: html_classes, **html_attributes, &block)
--- a/app/helpers/govuk_list_helper.rb
+++ b/app/helpers/govuk_list_helper.rb
@@ -1,24 +1,15 @@
 module GovukListHelper
   def govuk_list(array = nil, type: nil, spaced: nil, classes: nil, html_attributes: {}, &block)
+    type = type.to_s
+
+    fail "Unrecognised type for govuk_list - should be :bullet or :number or nil" unless type.in?(["bullet", "number", ""])
+
+    tag_type = (type == "number") ? "ol" : "ul"
+
     html_classes = ["#{brand}-list"]
-
-    case type.to_s.to_sym
-    when :bullet
-      html_classes << ["#{brand}-list--bullet"]
-      tag_type = "ul"
-    when :number
-      html_classes << ["#{brand}-list--number"]
-      tag_type = "ol"
-    when :""
-      tag_type = "ul"
-    else
-      raise "Unrecognised type for govuk_list - should be :bullet or :number or nil"
-    end
-
-    if spaced
-      html_classes << ["#{brand}-list--spaced"]
-    end
-
+    html_classes << "#{brand}-list--bullet" if type == "bullet"
+    html_classes << "#{brand}-list--number" if type == "number"
+    html_classes << "#{brand}-list--spaced" if spaced
     html_classes += Array.wrap(classes)
 
     if block_given?
--- a/app/helpers/govuk_list_helper.rb
+++ b/app/helpers/govuk_list_helper.rb
@@ -1,0 +1,36 @@
+module GovukListHelper
+  def govuk_list(array = nil, type: nil, spaced: nil, classes: nil, html_attributes: {}, &block)
+    html_classes = ["#{brand}-list"]
+
+    case type.to_s.to_sym
+    when :bullet
+      html_classes << ["#{brand}-list--bullet"]
+      tag_type = "ul"
+    when :number
+      html_classes << ["#{brand}-list--number"]
+      tag_type = "ol"
+    when :""
+      tag_type = "ul"
+    else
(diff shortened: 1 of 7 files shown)


app/helpers/govuk_list_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+  def govuk_list(array = nil, type: nil, spaced: nil, classes: nil, html_attributes: {}, &block)
+    html_classes = ["#{brand}-list"]
+
+    case type.to_s.to_sym
+    when :bullet
+      html_classes << ["#{brand}-list--bullet"]
+      tag_type = "ul"
+    when :number
+      html_classes << ["#{brand}-list--number"]
+      tag_type = "ol"
```
frankieroberto, line 11: Technically you _could_ have numbers for an ordered list, or bullets for an ordered list (the Design System doesn’t have a Nunjucks macro for this, but the classes would work) - however I don’t think there’s any real use case where this would make sense?

peteryates: Yeah, I'd agree. Would having these as separate named args `bullets: true` and `numbers: true` make it clearer? It could `fail` when both are true and gets rid of the need to do the symbol/string conversions.

app/helpers/govuk_list_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+      tag_type = "ul"
+    else
+      raise "Unrecognised type for govuk_list - should be :bullet or :number or nil"
+    end
+
+    if spaced
+      html_classes << ["#{brand}-list--spaced"]
+    end
+
+    html_classes += Array.wrap(classes)
```
peteryates, line 22: I think Rails' `#class_names` can make this a bit clearer, something along the lines of:

```suggestion
    type = type.to_s
    fail "Unrecognised type for govuk_list - should be :bullet or :number or nil" unless type.in?(["bullet", "number", ""])

    tag_type = (type == "number") ? "ol" : "ul"

    html_classes = class_names(
      "#{brand}-list",
      *classes,
      "#{brand}-list--bullet" => type.to_s == "bullet",
      "#{brand}-list--number" => type.to_s == "number",
      "#{brand}-list--spaced" => spaced
    )
```

app/helpers/govuk_list_helper.rb
frankieroberto: Hmm, interesting. I think that makes it less clear, at least to me? Having a single `type` argument more closely implies that it can only be bullets OR numbers?

app/helpers/govuk_list_helper.rb
frankieroberto: I’d not come across `class_name` before.

I had a quick go at this, but I think the custom classes (`*classes`) ought to appear after the `govuk-` ones as it can affect the CSS if they otherwise conflict, and `*classes` can't appear after the conditional ones for some reason.

I’ve used your suggestion to refactor the rest of this a bit though.

peteryates: Yeah, the arguments for class names is basically `*args, **kwargs` so you can't add an extra one on the end. CSS class order doesn't matter though - I think the worst that can happen is that someone might do `type: "number"` **and** `classes: "govuk-list--bullet"`, which might be confusing.

guide/lib/examples/list_helpers.rb
```diff
@@ -0,0 +1,30 @@
...
+        = govuk_list ["Delivery address.", "Payment.", "Confirmation."], type: :bullet
+      EXAMPLE
+    end
+
+    def govuk_list_spaced
+      <<~EXAMPLE
+        = tag.p "You will have to apply the reverse charge if you supply any of these services:"
+        = govuk_list type: :bullet, spaced: true do
+          = tag.li "constructing, altering, repairing, extending, demolishing or dismantling buildings or structures (whether permanent or not), including offshore installation services"
+          = tag.li "pipelines, reservoirs, water mains, wells, sewers, industrial plant and installations for purposes of land drainage, coast protection or defence"
```
peteryates, line 26: Glad you've included this example, shows max flexibility without having to remember class names.

peteryates (approved):
Looks good to me. I think `class_names` is neater but if you think the order matters I'm happy with how it is here.

app/helpers/govuk_list_helper.rb
peteryates: It's the order in the CSS that matters though, not the HTML, like here the second item should be yellow:

https://codepen.io/peteryates/pen/mdgerEQ

frankieroberto: You’re right, I remembered that completely wrong! 🤦‍♂️

frankieroberto: I've switched to using `class_names` now. 👍  One advantage it has is avoiding using `Array.wrap()` which is a Rails monkeypatch. Makes it easier if we were able to find a way to avoid the Rails dependency in future.

PR #521, opened 2024-03-13
Update tabs component to have a default title

### LINKED ISSUE #507: Default title for govuk_tabs
The `govuk_tabs` helper throws an exception if the `title` argument is omitted. The equivalent Nunjucks component [`govukTabs`](https://design-system.service.gov.uk/components/tabs/) defaults the title to "Contents" and doesn’t even include it in the examples.

I’m happy for us to require arguments where it makes sense to (Nunjucks doesn't support required arguments at all with the way that arguments are passed in as a object) – but maybe here we could default to "Contents" too?  Not sure how often you'd use another title?

See also #506

frankieroberto:
Set the default title to 'Contents' if not specified.

Fixes #507


peteryates (approved):
Nice 👌🏽

PR #530, opened 2024-04-05
Use nested hash data attributes in build_data_attributes method

### LINKED ISSUE #529: Gem can generate invalid duplicate data attributes on buttons
In some cases, the gem can generate two identical data attributes on a single button. 

## Steps to reproduce
Create a button with a data-module attribute using the Rails nested hash convention:
 
```ruby
<%= govuk_button_to 'A button', '#', data: { module: "special-js-button" } %>
```

observe that the generated HTML now contains two data-module atttributes:
```html
<button class="govuk-button" data-module="govuk-button" new_tab="false" data-module="special-js-button" type="submit">A button</button>
```
When this happens, browsers will discard the second data-module value as it is invalid. 

Note that if you use the full `data-module` attribute instead:
```ruby
<%= govuk_button_to 'A button', '#', 'data-module': "special-js-button" } %>
```

the generated HTML only contains our custom data-module atttribute:
```html
<button class="govuk-button" data-module="special-js-button" new_tab="false" type="submit">A button</button>
```

## Expected behaviour
I would expect to be able to overwrite the data-module attribute using the nested hash convention.

## Suspected cause
I think this is happening because of the way build_data_attributes works -  it uses the full `data-module` attribute rather than the nested hash:

```ruby
def build_data_attributes(data_module, prevent_double_click: nil)
    {
      "data-module": data_module,
      "data-prevent-double-click": prevent_double_click
    }.compact
  end
```

The `govuk_button_to` helper uses `deep_merge_html_attributes` which comes from the [x-govuk/html-attributes-utils](https://github.com/x-govuk/html-attributes-utils) gem. [Issue 1](https://github.com/x-govuk/html-attributes-utils/issues/1) in that repo says:

> This library, and both the [components](https://github.com/dfe-digital/govuk-components) and [form builder](https://github.com/dfe-digital/govuk-formbuilder) assume people are using the latter. Using the former might lead to unintended consequences.

So I think the solution here might be to use the nested hash convention in build_data_attributes:

```ruby
def build_data_attributes(data_module, prevent_double_click: nil)
    {
        data: {
          module: data_module,
          'prevent-double-click': prevent_double_click
        }.compact
    }
end
```

DavidBiddle:
Uses the nested hash method of declaring data attributes in build_data_attributes, as this is what `deep_merge_html_attributes` expects.

Fixes #529

### DIFF
--- a/app/helpers/govuk_link_helper.rb
+++ b/app/helpers/govuk_link_helper.rb
@@ -188,9 +188,11 @@ private
 
   def build_data_attributes(data_module, prevent_double_click: nil)
     {
-      "data-module": data_module,
-      "data-prevent-double-click": prevent_double_click
-    }.compact
+      data: {
+        module: data_module,
+        "prevent-double-click": prevent_double_click
+      }.compact
+    }
   end
 end
 
--- a/spec/helpers/govuk_link_helper_spec.rb
+++ b/spec/helpers/govuk_link_helper_spec.rb
@@ -181,10 +181,10 @@ RSpec.describe(GovukLinkHelper, type: 'helper') do
     end
 
     context "when there are custom attributes" do
-      let(:kwargs) { { lang: "en-GB", dir: "ltr" } }
+      let(:kwargs) { { lang: "en-GB", dir: "ltr", data: { module: "custom-module" } } }
 
       specify "the custom attributes are present on the link" do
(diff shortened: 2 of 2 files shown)


peteryates:
Ah, well caught and solved. Thank you, I'll get this merged and released soon.

peteryates (approved):
Looks great and works as expected.

PR #532, opened 2024-04-11
Allow tabs to have set ids

misaka:
The tab id attribute is currently based on the label given the tab. In some cases, however, this label may be dynamic, e.g. if the label includes the number of records included in that tab. Having an id that changes this way can get in the way of testing or potentially other functionality that may depend on the id being static.

### DIFF
--- a/app/components/govuk_component/tab_component.rb
+++ b/app/components/govuk_component/tab_component.rb
@@ -21,15 +21,16 @@ private
   class Tab < GovukComponent::Base
     attr_reader :label, :text
 
-    def initialize(label:, text: nil, classes: [], html_attributes: {})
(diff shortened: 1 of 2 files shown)


misaka:
This came out of work we were doing on our own project. I don't know how useful it is, np if it's not seen as worth adding, just thought I'd offer the contribution if it is useful.

frankieroberto (approved):
Makes sense to me. Good to have stable IDs in the URL fragments, eg for links and bookmarks.

peteryates (approved):
Yep, this makes sense to me. I might do a small follow up to tweak the method names too. Thanks @misaka

PR #533, opened 2024-04-16
Add custom tag to notification banner heading

### LINKED ISSUE #531: Notification banner: with_heading can't generate heading tags
The notification banner's `with_heading` method will always produce a div with the class `#{brand}-notification-banner__heading`. This is ideal for the case where the heading is the only content in the banner (because the content isn't actually a heading semantically, just large text):

```html
<div class="govuk-notification-banner" role="region" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
  <div class="govuk-notification-banner__header">
    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
      Important
    </h2>
  </div>
  <div class="govuk-notification-banner__content">
    <p class="govuk-notification-banner__heading">
      You have 7 days left to send your application.
      <a class="govuk-notification-banner__link" href="#">View application</a>.
    </p>
  </div>
</div>
```

However, if there's also other, non-heading content in the banner, we'd expect the heading to be a heading element, since it's semantically a heading for the content below:

```html
<div class="govuk-notification-banner govuk-notification-banner--success" role="alert" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
  <div class="govuk-notification-banner__header">
    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
      Success
    </h2>
  </div>
  <div class="govuk-notification-banner__content">
    <h3 class="govuk-notification-banner__heading">
      Training outcome recorded and trainee withdrawn
    </h3>
    <p class="govuk-body">Contact <a class="govuk-notification-banner__link" href="#">example@department.gov.uk</a> if you think there’s a problem.</p>
  </div>
</div>
```

(code examples taken from the [Design system docs for the notification banner component](https://design-system.service.gov.uk/components/notification-banner/)).

As things stand, if a user calls this component with both `with_heading` and a block, the generated code will fail WCAG 1.3.1 Info and Relationships (Level A), because the component would have a visible heading which isn't semantically marked up.

## Possible solutions
I can think of a few possible ways tro deal with this (there are probably others).

### 1. Leave things as they are and encourage people to use the block syntax
The govuk-frontend nunjucks macro lets you pass in an arbitrary block of HTML, which this implementation also allows. So users can manually pass the heading into the component:

```erb
<%= govuk_notification_banner(blah) do %>
  <h3 class="govuk-notification-banner__heading">
    Training outcome recorded and trainee withdrawn
  </h3>
  <p class="govuk-body">Contact <a class="govuk-notification-banner__link" href="#">example@department.gov.uk</a> if you think there’s a problem.</p>
<% end %>
```

This works, but I'm not sure it's intuitive for users - I think they'd probably expect (as our team did initially) `with_heading` to generate a heading element.

### 2. Get rid of the `with_heading` helper
The idea here would be to make the API more like the nunjucks macro - so users always have to pass in the markup for the heading, whether it's the fake div heading or a semantic one. This would probably be the simplest code change for this library, but it's annoying for anyone who's currently using `with_heading` and I'm sure there would be complexities around deprecation etc.

### 3. Make the `with_heading` helper generate the heading tag automatically
We could change the way the heading generation works so that it automatically uses a heading tag if the user has provided block content to the component, and uses a `div` if they haven't. This has the advantage of 'just working' without the user having to think about it, but on the other hand some users might be surprised by the change to the generated HTML. Presumably the heading tag would be `h#{title_heading_level + 1`, or `div` if `title_heading_level` is 6.

### 4. Add an optional `tag` parameter to `with_heading`
We could add an optional `tag` or `heading_level` parameter to the `with_heading` helper, so that the user can easily customise the heading:
```erb
<%= govuk_notification_banner(blah) do |banner| %>
  <% banner.with_heading(text: "Training outcome recorded and trainee withdrawn", tag: "h3")) %>
  <p class="govuk-body">Contact <a class="govuk-notification-banner__link" href="#">example@department.gov.uk</a> if you think there’s a problem.</p>
<% end %>
```
This would return `div` if not specified by the user. This has the advantage of not changing anyone's existing HTML, and is the most flexible, but it does require people to remember to set it.

DavidBiddle:
Adds an optional tag to the notification banner component's heading, so that the user can easily customise the heading. This makes it easier for users to create notification banners with the correct heading heirarchy.

Closes #531.

### DIFF
--- a/app/components/govuk_component/notification_banner_component.rb
+++ b/app/components/govuk_component/notification_banner_component.rb
@@ -31,18 +31,19 @@ class GovukComponent::NotificationBannerComponent < GovukComponent::Base
   end
 
   class Heading < GovukComponent::Base
-    attr_reader :text, :link_href, :link_text
+    attr_reader :text, :link_href, :link_text, :tag
 
-    def initialize(text: nil, link_text: nil, link_href: nil, classes: [], html_attributes: {})
+    def initialize(text: nil, link_text: nil, link_href: nil, tag: 'p', classes: [], html_attributes: {})
       @text      = text
       @link_text = link_text
       @link_href = link_href
+      @tag       = tag
 
       super(classes:, html_attributes:)
     end
 
     def call
-      tag.div(**html_attributes) do
+      content_tag(tag, **html_attributes) do
         if text.present?
           safe_join([text, link].compact, " ")
         else
--- a/spec/components/govuk_component/notification_banner_component_spec.rb
+++ b/spec/components/govuk_component/notification_banner_component_spec.rb
@@ -25,22 +25,25 @@ RSpec.describe(GovukComponent::NotificationBannerComponent, type: :component) do
         render_inline(described_class.new(**kwargs)) do |component|
           component.with_heading(**slot_kwargs)
           component.with_heading(text: 'More text here')
+          component.with_heading(tag: 'h3', text: "A heading with a custom tag")
           component.with_heading do
-            helper.tag.p('some special content')
+            helper.tag.span('some special content')
           end
         end
       end
 
       specify 'headings are rendered with content' do
         expect(rendered_content).to have_tag('div', with: { class: 'govuk-notification-banner__content' }) do
-          with_tag('div', with: { class: 'govuk-notification-banner__heading' }, text: /some text/) do
+          with_tag('p', with: { class: 'govuk-notification-banner__heading' }, text: /some text/) do
             with_tag('a', with: { class: 'govuk-notification-banner__link', href: '#look-at-me' }, text: 'With a link')
           end
 
-          with_tag('div', with: { class: 'govuk-notification-banner__heading' }, text: 'More text here')
+          with_tag('p', with: { class: 'govuk-notification-banner__heading' }, text: 'More text here')
 
-          with_tag('div', with: { class: 'govuk-notification-banner__heading' }) do
-            with_tag('p', text: 'some special content')
+          with_tag('h3', with: { class: 'govuk-notification-banner__heading' }, text: 'A heading with a custom tag')
+
+          with_tag('p', with: { class: 'govuk-notification-banner__heading' }) do
+            with_tag('span', text: 'some special content')
           end
         end
       end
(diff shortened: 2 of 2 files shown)


app/components/govuk_component/notification_banner_component.rb
```diff
@@ -31,18 +31,19 @@ def render?
...
   class Heading < GovukComponent::Base
-    attr_reader :text, :link_href, :link_text
+    attr_reader :text, :link_href, :link_text, :tag
 
-    def initialize(text: nil, link_text: nil, link_href: nil, classes: [], html_attributes: {})
+    def initialize(text: nil, link_text: nil, link_href: nil, tag: 'div', classes: [], html_attributes: {})
       @text      = text
       @link_text = link_text
       @link_href = link_href
+      @tag = tag
```
peteryates, line 40: Mind aligning the equals here please?

peteryates:
Sorry left a comment before realising it was still in draft! Looking spot on though 👌🏽

app/components/govuk_component/notification_banner_component.rb
```diff
@@ -31,18 +31,19 @@ def render?
   end
 
   class Heading < GovukComponent::Base
-    attr_reader :text, :link_href, :link_text
+    attr_reader :text, :link_href, :link_text, :tag
 
-    def initialize(text: nil, link_text: nil, link_href: nil, classes: [], html_attributes: {})
+    def initialize(text: nil, link_text: nil, link_href: nil, tag: 'div', classes: [], html_attributes: {})
```
frankieroberto, line 36: Should we default to `p` as [used in the Nunjucks version](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/notification-banner/template.njk#L43-L46) if you just specify `text`? 🤔

DavidBiddle: Yeah, that makes sense, will update!

DavidBiddle: @frankieroberto done, thanks for the suggestion 😄

frankieroberto (approved):
Looks good to me, thanks!

DavidBiddle:
Thanks @frankieroberto - I don't have permissions to merge on this repo, is that something you're able to do?

frankieroberto:
> Thanks @frankieroberto - I don't have permissions to merge on this repo, is that something you're able to do?

I can, but I’ll let @peteryates do a quick final check too.

DavidBiddle:
Thank you both for your help on this!

PR #540, opened 2024-06-05
Allow default summary list value text to be configured

### LINKED ISSUE #411: Add fallback text for summary list values
Sometimes across projects we end up with 'or' statements within views so if a attribute is missing we display some relevant text instead, like this:

```ruby
summary_list_row.with_value(text: object.stuff || "Not set")
```

It's a bit untidy and can get repetitive.

Perhaps if we allowed a `fallback_text` option that could have a default, we could remove some of this duplication and noise. When used on its own it might look like this:

```ruby
summary_list_row.with_value(text: object.stuff, fallback_text: "Not set")
```

But with a default it's much cleaner.

```ruby
# in config/components.rb
Govuk::Components.configure do |conf|
  default_summary_list_fallback_text: "Not set"
end

# in your view
summary_list_row.with_value(text: object.stuff)
```

Here, like we've done elsewhere, we could disable it on an ad hoc basis with `fallback_text: false`.

peteryates:
Being able to set the default fallback text for summary list values is convenient and saves people from peppering extra logic throughout the views.

The new configuration value, `default_summary_list_value_text`, defaults to `""` so the behaviour of the library won't change unexpectedly.

Fixes #411

### Final tasks

* [ ] add a paragraph (or maybe new section) to the guide explaining the behaviour and suggesting [better alternatives than fallback text](https://design-system.service.gov.uk/components/summary-list/#showing-missing-information) following @frankieroberto's suggestion

### DIFF
--- a/app/components/govuk_component/summary_list_component/value_component.rb
+++ b/app/components/govuk_component/summary_list_component/value_component.rb
@@ -18,6 +18,6 @@ private
   end
 
   def value_content
-    content || text || ""
+    content || text || config.default_summary_list_value_text
   end
 end
--- a/guide/content/components/summary-list.slim
+++ b/guide/content/components/summary-list.slim
@@ -43,7 +43,12 @@ p
   code: summary_list_with_missing_information) do
 
   markdown:
-    If a row has missing information, include a link to add the information within the value column instead of as an action.
+    If a row has missing information, include a link to add the information
+    within the value column instead of as an action.
+
+    When there is no link fallback text will be used. By default it's
+    a blank string but can be overridden by [configuring](/introduction/configuration/)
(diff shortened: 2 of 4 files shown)


frankieroberto:
Makes sense, but perhaps we should be nudging people towards this pattern instead: https://design-system.service.gov.uk/components/summary-list/#showing-missing-information - rather than having default text like "Not set"? 🤔

peteryates:
Yeah, I agree. I think this option and the behaviour can easily be explained in a couple of paragraphs in the guide. Still, adding a link requires effort so being able to configure a fallback (even one that defaults to `""`) is likely to have utility I think.

frankieroberto (approved):

PR #556, opened 2024-09-02
Service navigation

frankieroberto:
The service navigation component sits beneath the header at the top of the page and contains the service's name and it's primary navigation.

This implementation follows the upstream one very closely but has some additional functionality:

* it accepts a `current_path` argument
* when navigation items match `current_path` exactly they are marked as current
* for pages nested within a section, it accepts a `active_when` argument
* when navigation items begin with the active_when value they are marked as 'active'
* if a regular expression is provided for `active_when`, any positive match results in the node being marked 'active'


frankieroberto:
@peteryates feel free to close this PR and re-open one with your name on it if you like. I only opened it to trigger the deploy preview.  Annoying there's no way (yet) to transfer a PR between users.

PR #573, opened 2024-11-20
Add custom heading level to summary cards

agoldstone93:
This extends the summary card to allow for changing the heading level from the default `h2`

### DIFF
--- a/app/components/govuk_component/summary_list_component/card_component.rb
+++ b/app/components/govuk_component/summary_list_component/card_component.rb
@@ -1,11 +1,12 @@
 class GovukComponent::SummaryListComponent::CardComponent < GovukComponent::Base
-  attr_reader :title
+  attr_reader :title, :heading_level
 
   renders_many :actions
   renders_one :summary_list, "GovukComponent::SummaryListComponent"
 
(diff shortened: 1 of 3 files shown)


app/components/govuk_component/summary_list_component/card_component.rb
```diff
@@ -17,6 +18,10 @@ def default_attributes
     { class: "#{brand}-summary-card" }
   end
 
+  def heading_tag
```
peteryates, line 21: I think it might make sense to do a check here to make sure the provided level is a number from `1..6`. Something similar happens in [the accordion component](https://github.com/x-govuk/govuk-components/blob/main/app/components/govuk_component/accordion_component.rb#L35).

peteryates (commented):
Looks great overall. Left one really minor suggestion but will be happy to accept this and get it in the next release. Thank you.

app/components/govuk_component/summary_list_component/card_component.rb
agoldstone93: @peteryates Good shout. I've updated to include this

PR #590, opened 2025-05-09
Allow setting custom licence HTML in footer

stephencdaly:
There is already an attribute to set custom licence text, however this does not allow including links or a logo, meaning it's not possible to internatialise the footer without removing these items.

Allow passing in `licence_meta_html` as a block to be included instead of the default licence content.

If this change seems sensible, I can attempt to update the documentation.

### DIFF
--- a/app/components/govuk_component/footer_component.html.erb
+++ b/app/components/govuk_component/footer_component.html.erb
@@ -32,14 +32,14 @@
             </div>
           <% end %>
 
-          <% if meta_licence.nil? %>
+          <% if meta_licence_content.nil? %>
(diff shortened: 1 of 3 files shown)


peteryates:
Very sensible! Thanks - looks great :slightly_smiling_face:

stephencdaly:
Thanks @peteryates! I've tried adding instructions for using this to the guide, but I'm not familiar with Slim so maybe it could be improved

peteryates:
The guide is out of action at the moment anyway because of #591 :disappointed: 

Just taking a look at this.

peteryates (approved):
Looks great. Thank you for contributing :slightly_smiling_face:

stephencdaly:
@peteryates is this good to be merged?

stephencdaly:
I don't think I have the ability to merge

PR #597, opened 2025-06-24
Add the inverse variant of service navigation

peteryates:
This adds support for the [inverse variant](https://govuk-frontend-review.herokuapp.com/components/service-navigation#heading-inverse) of the service nav component.

## Preview

image

## Notes

Looks a bit weird when there's a service name on it too. I'd expect the service name to be white here :thinking: 

Did I miss something in the markup?

Screenshot From 2025-06-24 22-56-41


peteryates:
The dark service name was fixed in alphagov/govuk-frontend#6046 so this is good to go.

PR #598, opened 2025-06-25
Use inverse service navigation for homepage of guide

paulrobertlloyd:
Am inferring if the inverse class should be used based on the presence of a page title, do you think this’ll work @peteryates?

Screenshot of section of home page.


peteryates:
Looks good, I tweaked it so the logic's in a helper.

PR #600, opened 2025-06-26
Use correct class names in cell widths

ethax-ross:
We are using the short/convenience width name as the class for cell widths, so:

```
row.with_cell(text: 'Name', width: 'one-third')
```

Will render out as:

```
<th class="govuk-table__header one-third" scope="col">Name</th>
```

As a result the width is not correctly applied; we need to use the fully qualified class name from GOV UK frotend.

Update `CellComponent` to convert to the correct class name, for example `govuk-!-width-one-third`.

Fix an issue with the specs where Nokogiri throws an error due to trying to run a CSS selectcor with an `!` in it; this needs to be escaped.

As a workaround you can pass in the fully qualified class name to the `width`, so `width: "govuk-!-width-one-third"`.

### DIFF
--- a/app/components/govuk_component/table_component/cell_component.rb
+++ b/app/components/govuk_component/table_component/cell_component.rb
@@ -4,17 +4,6 @@ class GovukComponent::TableComponent::CellComponent < GovukComponent::Base
   alias_method :numeric?, :numeric
   alias_method :header?, :header
 
-  def self.widths
-    {
-      "full"           => "#{brand}-!-width-full",
-      "three-quarters" => "#{brand}-!-width-three-quarters",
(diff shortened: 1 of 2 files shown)


app/components/govuk_component/table_component/cell_component.rb
```diff
@@ -75,7 +75,7 @@ def default_classes
     class_names(
       "#{brand}-table__#{class_suffix}",
       "#{brand}-table__#{class_suffix}--numeric" => numeric?,
-      width => width?,
+      self.class.widths[width] => width?,
```
peteryates, line 78: If the widths are moved to a constant it might be neater to do this in the initializer with

```
@width = WIDTHS.fetch(width, nil)
```

PR #602, opened 2025-06-30
Update guide to use scoped component previews (and other fixes)

paulrobertlloyd:
I started by updating `@x-govuk/govuk-prototype-components` to v5.0 and adding the now required `x-govuk-masthead--inverse` modifier class…

…in doing so I upgraded Sass, and got all of its ‘fun’ deprecation warnings. So I tried to reorganise the files to use the Sass module system (as much as is possible given `govuk-frontend` doesn’t full support this yet).

…which made me realise I could add `pkg:` importing, but that required updating the `dart_sass` Nanoc filter to enable passing options such as `pkg-importer` to it. Once I’d done that…

…the prose styles used `@extend`, which doesn’t work great with Sass modules. But I realised these styles wouldn’t be needed if pages used `govuk_markdown` as the markdown renderer, so updated the config for Slim templates.

…and then I proceeded to update the places which should use the Markdown filter  instead of Slim element blocks

…but why can’t some pages just use Markdown? So I added a Markdown Nanoc filter

…and converted a few pages to Markdown.

I’ve also been thinking about the component previews, and how they display using the wrong font and brand colour. Earlier in the day I wondered if moving component HTML into a web component would mean the shadow DOM could use and scope a separate set of styles. So I tried this and it worked!

…but then I needed to faff around with getting the corresponding JavaScript to work, and initialise within the scope of an individual components shadow DOM. With a bit of help from a chatbot, I figured out a way through.

So, I kinda got carried away with myself, whoops! 🫠

Creating this PR to preview the changes and get feedback on this approach and my litany of tweaks and changes!


paulrobertlloyd:
Fixed tabs not working, needed to import and manually instantiate them.

paulrobertlloyd:
@peteryates Horribly close to the truth!

guide/content/stylesheets/preview.scss
```diff
@@ -75,17 +75,11 @@
...
-
-  margin-right: 0;
-  fill: tomato;
-  vertical-align: top;
+  fill: #00ffe0;
 }
 
 .app-header__logotext {
-  color: tomato;
+  color: #00ffe0;
```
peteryates, line 82: :tomato:

peteryates (approved):
Looks great, the scoped previews are much clearer :ok_hand:

PR #613, opened 2025-09-01
Defer helper usage in Footer, Action and Accordion components for VC4

avinhurry:
## Summary

This PR updates `govuk-components` to be compatible with **ViewComponent v4** by deferring all helper usage until render time.

## Changes

- **FooterComponent**
  - Store raw `meta_items` in `initialize`.
  - Defer building meta links until `before_render`.
  - Add type check in `initialize` to raise if `meta_items` is not a hash/array (keeps existing spec passing).

- **SummaryList::ActionComponent**
  - Move `govuk_link_classes` call out of `default_attributes`.
  - Compute link classes in `call` instead, where helpers are available.
  - `default_attributes` now returns `{}`.

- **AccordionComponent**
  - Defer reading `html_attributes[:id]` until `before_render`.
  - Prevents accessing view context–dependent state in `initialize`.

## Why

ViewComponent v4 removed the default initializer and no longer provides the view context during `initialize` or `default_attributes`. Any use of helpers in those methods raises errors.

These changes ensure helpers (`govuk_link_classes`, `govuk_footer_link_to`, etc.) are only called during `before_render` or `call`, where the view context is available.

## References

- [ViewComponent v4 changelog](https://github.com/ViewComponent/view_component/releases/tag/v4.0.0)

  > Remove default initializer from ViewComponent::Base.  
  > Remove dependency on ActionView::Base, eliminating the need for capture compatibility patch.  
  > In some edge cases, this change may require switching to use the `helpers` proxy.

### DIFF
--- a/app/components/govuk_component/accordion_component.rb
+++ b/app/components/govuk_component/accordion_component.rb
@@ -16,9 +16,13 @@ class GovukComponent::AccordionComponent < GovukComponent::Base
 
   def initialize(heading_level: 2, classes: [], html_attributes: {})
     @heading_level = heading_tag(heading_level)
-    @accordion_id  = html_attributes[:id]
 
     super(classes:, html_attributes:)
+  end
+
+  def before_render
+    super
+    @accordion_id = html_attributes[:id]
   end
 
   def call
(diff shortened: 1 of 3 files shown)


avinhurry:
Closing as it has been superseded by: https://github.com/x-govuk/govuk-components/pull/615

PR #626, opened 2026-01-31
Update header component for govuk-frontend 6.0.0

peteryates:
The upstream header component received quite a lot of attention in version 6.0.0. The headline changes are:

* service name and navigation links have been removed
* the component is no longer rendered in a `<header>` element, it's now in a div and should be placed in a `<header>` by the user - this allows it to sit alongside the service nav and phase banner
* the GOV.UK link now defaults to https://www.gov.uk, not the service's root
* the GOV.UK link's class name has been simplified

This change updates the existing header component to match the new functionality.

## Removal of service name and navigation links

This functionality has been removed entirely.

https://github.com/alphagov/govuk-frontend/pull/6423

This _might_ cause some upset as plenty of teams do have links in the main header for things like signing-in and out of services. I think we can find a workable compromise by using the [content accessor](https://viewcomponent.org/guide/getting-started.html#implementation), but teams will need to supply their own styling etc.

## Rendering of service navigation and phase banner in the header

The component has been changed so the GOV.UK header sits within `<div>` inside the `<header>` element (rather than the `<header>` element being the container).

The upstream change is described as follows:

> We've changed the page structure around the header to separate the header element `<header>` from the GOV.UK header component. This lets you include other components, such as the Service navigation and Phase banner components, within the header element of each page.
>
> You’ll need to follow different instructions to upgrade, depending on how you create the page headers in your service. In all cases, you should make sure your page includes a single `<header>` element after you’ve made the changes.

I think this is due to a limitation of Nunjucks more than anything, and `govuk_header` **not** rendering a `<header>` is likely going to be more confusing.

So, now the service navigation and phase banner are available as slots:

```erb
<%=
  govuk_header do |header|
    header.with_service_navigation(
      service_name: "My fancy service",
      service_url: "www.fancy-service.com",
      navigation_items: [...]
    )
    header.with_phase_banner(
      tag: { text: 'Alpha' },
      text: "This is a new service. Be nice"
    )
  end
%>
```

## Changing the default homepage URL from `/` to `https://www.gov.uk`

This was changed in govuk-frontend and updated here to match. As before, it can be overridden if necessary using the `default_header_homepage_url` config setting.

### DIFF
--- a/guide/content/stylesheets/preview.scss
+++ b/guide/content/stylesheets/preview.scss
@@ -67,6 +67,19 @@
   }
 }
 
+.govuk-header.with-extra-content > .govuk-header__container {
+  display: flex;
+
+  .govuk-header__logo {
+    flex-grow: 1;
+  }
+
+  .account-options {
+    align-content: center;
+    font-weight: bold;
+  }
+}
+
 // Example custom logo used in documentation
 .app-header__logotype {
   fill: #00ffe0;
--- a/app/components/govuk_component/header_component.html.erb
+++ b/app/components/govuk_component/header_component.html.erb
@@ -36,6 +36,7 @@
           <%= product_name %>
         <% end %>
       </div>
+      <%= content %>
     <% end %>
   <% end %>
   <%= service_navigation %>
--- a/guide/content/components/header.slim
+++ b/guide/content/components/header.slim
@@ -40,4 +40,11 @@ markdown:
     [service navigation](/components/service-navigation) and a [phase banner](/components/service-navigation) alongside using
     the provided slots.
 
+== render('/partials/example.*',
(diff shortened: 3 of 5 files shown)


frankieroberto:
@peteryates I still live in hope that the One Login header bits will get upstreamed into govuk-frontend proper... 🤞

guide/content/stylesheets/preview.scss
```diff
@@ -95,4 +95,17 @@
 // height
 .app-scoped-preview {
   contain: layout;
+
+  .app-header > .govuk-header__container {
```
peteryates: I suspect this doesn't need to be scoped to the preview, I'll move it out.

Issue #641, opened 2026-05-15
Update footer component to use `div`


DavidBiddle:
[govuk-frontend v6](https://github.com/alphagov/govuk-frontend/releases/tag/v6.0.0) includes a [change to the footer component markup](https://github.com/alphagov/govuk-frontend/pull/6537) which decouples the HTML `footer` element from the GOV.UK footer component.

```html
<footer class="govuk-template__footer">
  <div class="govuk-footer">
    [...]
  </div>
</footer>
``` 

I think the simplest way to deal with this would be to:
- update the footer component to use `div` instead of `footer`
- add some content to the guidance around wrapping the component in a footer element

There's a disadvantage here as it means anyone who updates without checking the guidance will be left with a `<footer>`-less page. The alternative would be rendering the `<footer>` element as part of the component and adding slots to the component for other footer content, but I think this would diverge a bit from the Nunjucks API in a way that is probably unexpected for users.

peteryates:
Thanks @DavidBiddle, I didn't notice this had changed too!

I did the corresponding header change in #626 and there did opt to have the header component render the `<header>` element. Removing it felt unintuitive, and more a limitation of Nunjucks that we can overcome nicely with slots.

Is there a common need for rendering the footer anywhere but in a `<footer>` element? 🤔

peteryates:
My corresponding change was missing the removal of `role: 'contentinfo'` 🤦‍♂️  This is hopefully now fully resolved though in v6.3.1

PR #642, opened 2026-05-15
Update footer to use V6 HTML

DavidBiddle:
Decouples the footer component from the HTML footer element, following the [corresponding change in govuk-frontend v6](https://github.com/alphagov/govuk-frontend/pull/6537).

Implements #641

### DIFF
--- a/app/components/govuk_component/footer_component.html.erb
+++ b/app/components/govuk_component/footer_component.html.erb
@@ -1,4 +1,4 @@
-<%= tag.footer(role: 'contentinfo', **html_attributes) do %>
+<%= tag.div(**html_attributes) do %>
   <%= tag.div(**container_html_attributes) do %>
     <%= tag.svg(class: "#{brand}-footer__crown", focusable: "false", role: "presentation", xmlns: "http://www.w3.org/2000/svg", viewBox: "0 0 64 60", height: "30", width: "32", fill: "currentColor") do %>
       <g>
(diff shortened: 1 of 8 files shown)


app/components/govuk_component/footer_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= tag.footer(role: 'contentinfo', **html_attributes) do %>
+<%= tag.div(**html_attributes) do %>
```
peteryates, line 1: I think having to remember to render the `govuk_footer` inside a `<footer>` element is a bit unintuitive and a backwards step, even if the upstream implementation does it. Plus, there's a good chance teams will upgrade without reading the notes and will end up footerless.

I'd rather keep the `<footer>` element in the template, preferably with similar behaviour to the header component.

So, if we render a footer and pass a block of HTML, that gets rendered inside the `<footer>` element but above the GOV.UK footer component (with `<%= content %>`).

```erb
<%= govuk_footer do %>
  <%= govuk_phase_banner(tag: "Alpha", text: "Hi") %>
<% end %>
```

which renders:

```html
<footer>
  <div class="govuk-phase-banner">...</div>
  <div class="govuk-footer">...</div>
</footer>
```

I suspect rendering content above the footer is going to be much more common than wanting to render content beneath it, but we could always add another slot for content below if the need arises.

DavidBiddle: Yeah, that all sounds fair - didn't realise you'd already made a similar header change so consistency makes sense to me. I'll close this PR!

PR #653, opened 2026-07-22
Rename footer attributes for consistency with headers

avinhurry:
## Summary

Rename `footer_attributes` to `footer_html_attributes` on the footer component to be consistent with the header components.

> [!IMPORTANT]
> This is a **breaking API change**, but there are currently no known usages of `footer_attributes`.


peteryates (approved):
Looks great, thank you!

PR #666, opened 2026-09-16
Allow overriding footer text with locales

agoldstone93:
Allow passing in locales to the govuk_footer so that the default text can easily be translated into another language


app/components/govuk_component/footer_component.rb
```diff
@@ -16,7 +16,7 @@ def initialize(
     classes: [],
     container_classes: [],
     container_html_attributes: {},
-    copyright_text: config.default_footer_copyright_text,
+    copyright_text: nil,
```
peteryates: I wonder if leaving the default here would make more sense, and then we can allow translations by explicitly setting the default value to `nil`. That would avoid the need for the extra conditional in `#copyright`.

app/components/govuk_component/footer_component.rb
```diff
@@ -88,12 +88,23 @@ def build_meta_links(links)
   end
 
   def default_licence
```
peteryates: Probably worth renaming `default_licence` to `open_government_licence` to make it clearer.

peteryates (commented):
This looks good to me. Just left a couple of thoughts.

app/components/govuk_component/footer_component.rb
agoldstone93: @peteryates Do you mean by passing in `copyright_text` into `govuk_footer` in the downstream application?

agoldstone93: @peteryates I've pushed some changes. Is that what you meant?

app/components/govuk_component/footer_component.rb
agoldstone93: @peteryates Good shout. I've updated it

app/components/govuk_component/footer_component.rb
peteryates: Thanks @agoldstone93.

Yeah I think this is looking good now.

I'm starting to wonder whether my original monolingual approach is going to make this unnecessarily complicated.

Perhaps moving all the default text from the config to a locale file will be cleaner. Then we only need to see if an override was passed in and fall back if not. :thinking:

I guess the difficulty with that is if someone wants to omit an arg (which you'd usually do with `thing: nil` and we end up then populating it with a translated string, it'd be unclear.

